PR #167, opened 2015-09-04
Fixed issue with exclusion lists whenever there is more than one type of molecule

jkrajniak:


junghans:
Looks ok to me!

@ruehle : any comments?

PR #173, opened 2015-11-14
H5MD: moved from C++ interface to C interface

jkrajniak:
This is done because on many HPC systems HDF5 library is only compiled with C API. This is because it is not possible (at least for HDF5 1.8.16) to compile library both with CXX interface and with support of parallel operations.


junghans:
Thanks!

PR #174, opened 2015-11-19
updated comment to settings xml option

jkrajniak:
Because we had to look into source code to figure out why this is not passed to csg_stat correctly.


junghans:
Thanks!

junghans:
Added that fix to the stable branch as df45dfc1b066b7103eba1eb485e29e2a12959558, too.

PR #176, opened 2015-11-22
Added reading of beads and bonded terms in XMLTopologyReader

jkrajniak:
Modifications:
- switched to XML parser based on Property tree
- new section **molecule** under `<molecules>` that stores `<bead>`

``` xml
    <molecules>
        <molecule name="BUT" nmols="4000" nbeads="4">
            <bead name="C1" type="C" mass="15.035" q="0.0" />
            <bead name="C2" type="C" mass="14.028" q="0.0" />
            <bead name="C3" type="C" mass="14.028" q="0.0" />
            <bead name="C4" type="C" mass="15.035" q="0.0" />                                                                                                             
        </molecule>
    </molecules>
```
- new section `<bonded>` to store bonded terms

``` xml
<bonded>
    <bond>
        <name>bond1</name>
        <beads>
            BUT:C1 BUT:C2
        </beads>
    </bond>
</bonded>
```

Example of topology for butane (UA) molecules: [topol.xml](https://gist.github.com/MrTheodor/4d6bba5ffb867146d265)


junghans:
@ruehle @symashayak : This is a big change.

My understand is that you want to have a more flexible molecule definition in topol.xml, right? And currently things can only be renamed from a base topology, which not xml, right?.Additionally bonds have to be defined in the mapping file, which is a bit awkward.

This needs to be documented in the manual and a new file `topol.xml` in `share/xml`.

As you comment is changed, did you find the bug in `csg_stat` not using the exclusions correctly?

jkrajniak:
I still work on that, but I want to share it to discuss the approach before I write more code ;-) The exclusions seem to work correctly, I missed something. 

The idea is to get rid of using .gro, .pdb as a base and have at least partially description of the system in single file. Recently we had to play with lammps and we had issue with naming of atoms in the molecules, so then some mapping could be done. Lammps does not have anything like that in dump file and xml can only do rename of molecules. In the end we use .tpr.

The bonded section might seems redundant but if one wants to use csg_stat without mapping, he still has to write 1:1 mapping with cg_bonded part to remove bonded terms from rdf.

junghans:
How is the status?

jkrajniak:
It's still on my todo list

junghans:
I think that would be a cool feature to have, so let me know if there is some, even limited, version to merge.

junghans:
Is this ready to merge?

src/libcsg/modules/io/xmltopologyreader.cc
```diff
@@ -24,152 +24,319 @@
...
-            throw runtime_error("invalid box tag - missing xx, yy or zz");
-        try {
-           m[0][0] = boost::lexical_cast<double>(xx);
-        } catch(boost::bad_lexical_cast &) {
-           throw std::runtime_error("Cannot convert xx='"+ xx + "' to double");
+void XMLTopologyReader::ParseBox(Property &p) {
+    matrix m;
+    m[0][0] = p.getAttribute<double>("xx");
+    m[1][1] = p.getAttribute<double>("yy");
+    m[2][2] = p.getAttribute<double>("zz");
```
junghans: Are the error messages still clear if one of them is missing and/or is not a double?

junghans: Forget it, it works!

jkrajniak:
Huh what a mess with all those commits from merge...

anyway, I noticed after looking on @symashayak data is that whenever <clear/> is used the molecule list is clear but not the bead list and second time beads are added.
(ref: https://groups.google.com/forum/#!topic/votca/vZonpPE7kG0)
I am not entirely sure if <clear/> should also clean bead list as the original functionality will be broken. Instead, if base="" is defined then it is not possible to define own beads.

junghans:
You have rebase all commit in master, a merge would make every commit show up twice in the history, so I cannot merge this in its current state.

You could either use `git rebase -i master` and pick important commits related to the xmltopoloy reader. Or simpler use the patch from `git diff HEAD..master` as a single commit on a new branch.

junghans:
And it is ok to break the original functionally if this makes the xml reader cleaner. We can always re-implement it later if needed.

jkrajniak:
and I'm back...

Tested:
csg_map --top topol.xml --trj spce.dump --out spce2.gro --no-map
and also with mapping file, it seems to be okey. 

Maybe @symashayak could test on some trajectory from lammps;

junghans:
Looks good to me! @symashayak can you test that?

symashayak:
Hi,

It seems to have worked for me too. Especially csg_dump output seems to be
correct.

But the water-water RDF I get with csg_stat with mapping has some errors.
This could be due to some error I may have done generating LAMMPS
trajectory.

So I am debugging that issue. Will update as soon as I am done testing.

Thanks,
Sikandar

symashayak:
It works for me too! RDF from csg_stat with mapping seems correct.

PR #178, opened 2015-12-19
csg_boltzmann: re-implemented --excl option (fixes #177)

### LINKED ISSUE #177: csg_boltzmann: --excl option is broken/not implemented anymore
In 6991f1e4a308f23f546181fed57a70d8ba286745 the `--excl` option of `csg_boltzmann` got broken. @ruehle drafted up some come [here](https://github.com/votca/csg/commit/6991f1e4a308f23f546181fed57a70d8ba286745#diff-96f1ebad49076ca88c838cbdef42b6beR100), but it still needs some work!

@votca/csg: Should we fix it or drop the `--excl` option?

@MrTheodor: Have you used that option before?

junghans:
I don't like the N^2 loop, but better than nothing.

@ruehle, @MrTheodor : review please!

### DIFF
--- a/src/csg_boltzmann/main.cc
+++ b/src/csg_boltzmann/main.cc
@@ -48,6 +48,7 @@ public:
     bool DoMapping() { return true; }
 
     void Initialize();
+    bool EvaluateOptions();
     void Run();
 
     void InteractiveMode();
@@ -63,9 +64,18 @@ void CsgBoltzmann::Initialize()
 {
     CsgApplication::Initialize();
     AddProgramOptions("Special options")
-        ("excl", boost::program_options::value<string>(), "write exclusion list to file");
+        ("excl", boost::program_options::value<string>(), "write atomistic exclusion list to file");
 
     AddObserver(&_bs);
+}
+
+bool CsgBoltzmann::EvaluateOptions()
+{
+        CsgApplication::EvaluateOptions();
+        if (OptionsMap().count("excl")) {
+            CheckRequired("cg", "excl options needs a mapping file");
+	}
+        return true;
 }
 
 bool CsgBoltzmann::EvaluateTopology(Topology *top, Topology *top_ref)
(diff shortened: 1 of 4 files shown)


jkrajniak:
Hi,

For --no-map it causes `Segmentation Fault`:

```
csg_boltzmann --top topol_new.xml --excl ooo --no-map

Program received signal SIGSEGV, Segmentation fault.
0x000000000040d8bd in CsgBoltzmann::EvaluateTopology(votca::csg::Topology*, votca::csg::Topology*) ()
```

I am not exactly sure how this should work, I tried on a very simple example of butane and it generates such excl list:

```
# atomistic: BUT cg: BUT cgmap: map.xml
1 3 4
2 3 4
```

I would expect that it will only generate excl for CG molecules so:

```
1 2
```

Test files:
https://gist.github.com/MrTheodor/da3f6de77e1143934708

junghans:
Hmm, isn't the `--excl` option to write exclusions to be used for the atomistic simulation?

I am not sure if it makes sense to combine it with `--no-map`, but let me check what 1.2.4 was doing.

junghans:
Ok, getting same results as with v1.2.4. 

There were a couple of things wrong in the new exclusion list scheme, too, pointer comparison is dangerous ;-)

In v1.2.4 in `--excl` + `--no-map` segfaults, too, so I disable that combination as it doesn't make much sense anyhow.

src/csg_boltzmann/main.cc
```diff
@@ -63,11 +64,20 @@ void CsgBoltzmann::Initialize()
...
+        ("excl", boost::program_options::value<string>(), "write atomistic exclusion list to file");
 
     AddObserver(&_bs);
 }
 
+bool CsgBoltzmann::EvaluateOptions()
+{
+        CsgApplication::EvaluateOptions();
+        if (OptionsMap().count("excl")) {
+            CheckRequired("map", "excl options needs a mapping file");
```
jkrajniak: Great! I think that it should be "cg" instead of "map".

junghans: Good point, fixed!

PR #180, opened 2016-05-20
Added scaling option in optimizer_target_density.sh

t3rodrig:
A portion of settings.xml file:

``` xml
<cg>
     .
     .
     .
      <lammps>
        <!-- name of the table for lammps run -->
        <table>CG_CG.pot</table>
        <table_begin>2.0</table_begin>
        <table_bins>0.025</table_bins>
      </lammps>
      <optimizer>
        .
        .
        .
        <density>
          <target>CG-CG.density.tgt</target>
          <min>0</min>
          <max>80</max>
          <step>0.5</step>
          <axis>z</axis>
          <!-- Scaling factor for density -->
          <scale>1.66054</scale>
        </density>
        </optimizer>
     .
     .
     .
</cg>
```

### DIFF
--- a/share/scripts/inverse/optimizer_target_density.sh
+++ b/share/scripts/inverse/optimizer_target_density.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 #
-# Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
+# Copyright 2009-2016 The VOTCA Development Team (http://www.votca.org)
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


junghans:
Looks good to me.

Just so that I understand the background, your scaling only acts on the current density and not the target one, right? And why are you not using the `target_weights` to scale the difference?

junghans:
And don't forget to update the copyright year in `*.sh`.

t3rodrig:
Yes, I am only scaling the current density and not the target one. My target density is in g/cm^3, and I am using "units real" in lammps input file. An scaling factor is needed in order to get the right units of the current density profile [grams*mole^-1/A^3 -> g/cm^3]. Maybe I am wrong, but `target_weights` has a different use, or not?

junghans:
Yeah, I was just thinking if you give the target density (e.g. `CG-CG.density.tgt`) in same units as what csg_stat calculated, you don't need an extra scale, however, I am fine with adding it!

PR #183, opened 2016-08-20
Remove assertion on IBond, IAngle, IDihedral (reverts partially commit: votca/csg@f7b149b5dd)

jkrajniak:
The assertion will cause error whenever the bond contains more than two beads (and angles more than 3) so the following fragment of mapping file isn't anymore correct:

```
    <cg_bonded>
        <bond>
            <name>bond_b0</name>
            <beads>
                A1 B1
                B1 A2
            </beads>
        </bond>
```

this also fails:
https://github.com/votca/csg-tutorials/blob/master/hexane/atomistic/hexane.xml#L29

csg_map: /data/leuven/307/vsc30783/software/votca-mrtheodor/src/csg/include/votca/csg/interaction.h:92: votca::csg::IBond::IBond(std::list<int, std::allocator<int>> &): Assertion `beads.size()<2' failed.

### DIFF
--- a/include/votca/csg/interaction.h
+++ b/include/votca/csg/interaction.h
@@ -89,7 +89,7 @@ public:
         { _beads.resize(2); _beads[0] = bead1; _beads[1] = bead2; }
 
     IBond(list<int> &beads)
-        { assert(beads.size()<2); _beads.resize(2); for(int i=0; i<2; ++i) { _beads[i] = beads.front(); beads.pop_front(); }}
+        { _beads.resize(2); for(int i=0; i<2; ++i) { _beads[i] = beads.front(); beads.pop_front(); }}
     double EvaluateVar(const Topology &top);
     vec Grad(const Topology &top, int bead);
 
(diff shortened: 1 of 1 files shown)


junghans:
Also see f7b149b5dd6106b7f1ee81a3b0878c0c5558359f

junghans:
I was able to reproduce this:

```
cmake -DCMAKE_BUILD_TYPE=Debug ..
csg_map --top topol.tpr --cg hexane.xml --trj conf.gro --out cg.gro
```

junghans:
Isn't the `assert()` just the wrong way around?

jkrajniak:
Looks fine or maybe directly add also checkup if the whole list is correct:
`assert(beads.size() >= 2 && beads.size() % 2 == 0)`

Actually this constructor is only used in one place:
src/libcsg/cgmoleculedef.cc

where there is already check up if the bead number is divided by 2, 3, 4

junghans:
I didn't do the check on the whole list there in case the list would be more general (in the future) and contains e.g. 1 bond and 1 angle.

PR #201, opened 2017-02-17
Pressure Correction for LAMMPS

jdmoore2004:
These files should allow LAMMPS to use pressure correction.  It has been tested for the simple pressure correction.


junghans (approved):

share/scripts/inverse/calc_pressure_lammps.sh
```diff
@@ -0,0 +1,37 @@
+#! /bin/bash
+#
+# Copyright 2009-2015 The VOTCA Development Team (http://www.votca.org)
```
junghans, line 3: The year is wrong, but I will fix that later.

junghans (approved):

PR #207, opened 2017-04-18
xmltopologyreader: add resid tag

junghans:
/cc @razziel89

### DIFF
--- a/src/libcsg/modules/io/xmltopologyreader.cc
+++ b/src/libcsg/modules/io/xmltopologyreader.cc
@@ -119,11 +119,13 @@ void XMLTopologyReader::ParseMolecules(Property &p) {
 
 void XMLTopologyReader::ParseMolecule(Property &p, string molname, int nbeads, int nmols) {
     vector<XMLBead*> xmlBeads;
+    vector<int> xmlResidues;
     for (Property::iterator it=p.begin(); it != p.end(); ++it) {
         if (it->name() == "bead") {
             string atname = it->getAttribute<string>("name");
             string attype = it->getAttribute<string>("type");
             double atmass, atq;
+            int resid;
             try {
                 atmass = it->getAttribute<double>("mass");
             } catch (runtime_error) {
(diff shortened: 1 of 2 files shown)


junghans:
An update of the manual, somewhere around here:
https://github.com/votca/csg-manual/blob/master/inputfiles.tex#L84
would be nice.

Plus a line here:
https://github.com/votca/csg/blob/master/share/xml/topol.xml#L16
is needed, I guess.

RaZziel:
Ah, you found an old account. I didn't know it was still active. I will use my current account instead and delete this one.

junghans:
@razziel89: Thanks for the documentation.

@MrTheodor: please review.

src/libcsg/modules/io/xmltopologyreader.cc
```diff
@@ -134,25 +136,58 @@ void XMLTopologyReader::ParseMolecule(Property &p, string molname, int nbeads, i
             } catch (runtime_error) {
                 atq = 0.0;
             }
+            try {
+                resid = it->getAttribute<int>("resid");
+                if (resid<=0)
+                    throw std::runtime_error(
```
jkrajniak: This runtime_error will be catched by the "catch" below and the code will proceed with resid=-1; Is it an intention?

src/libcsg/modules/io/xmltopologyreader.cc
```diff
@@ -134,25 +136,58 @@ void XMLTopologyReader::ParseMolecule(Property &p, string molname, int nbeads, i
...
+                        "Residue count for beads in topology.molecules.molecule has to be greater than zero");
+            } catch (runtime_error) {
+                resid = -1;
+            }
+            if (!xmlResidues.empty()){
```
jkrajniak: Minor, a bit of mixed coding styles, missing spaces before brackets etc..

src/libcsg/modules/io/xmltopologyreader.cc
razziel89: Thanks for spotting this! It was not intentional and I replaced most of the `std::runtime_error` I added by `std::invalid_argument` as this won't be caught and fits better IMHO. The attached diff `small.txt` should hopefully alleviate this and the coding style issues as well. The attached diff `total.txt` is a diff with respect to the original. Using diffs is quite tedious. Should I create another pull request (I probably should have created one straight away)?

[small.txt](https://github.com/votca/csg/files/940997/small.txt)
[total.txt](https://github.com/votca/csg/files/940999/total.txt)

jkrajniak: The easiest way is to commit changes to the branch from where the pull request is made.

razziel89: I know this. Unfortunately, I did not create this pull request but @junghans did from a small diff that I posted on the forum.

junghans: You can just make another pull request with `xml_residue` being the base branch, but I included the changes quickly this time.

junghans:
@MrTheodor: happy?

src/libcsg/modules/io/xmltopologyreader.cc
razziel89: Ah, thank you both very much for all your help and patience!

junghans: No worries, the first couple of pull requests are the hardest!

jkrajniak (approved):
LGTM

PR #217, opened 2017-08-13
CsgSimplex.pm: fix double def of get_convergence_value

junghans:


abrukhno:
Hi, not much to review, but yes it was a double definition (survived since version 1.3 at least, apparently).
Andrey

Issue #210, opened 2017-06-20
-tableb option in run_gromacs.sh


schererc:
At the moment I am cleaning up the csg tutorials. I figured out that at least with my installed Gromacs version (5.1) the option -tableb table_a1.xvg table_b1.xvg causes an error in the simulation step if I run any ibi calculation with tabulated bonded interactions (wether they are updated or not). If I run without the -tableb option it works (tables have to have the same number as in the topol.top file). If I simply remove the option, it works (at least for Gromacs 5.1)

junghans:
I am guessing hit by:
<https://groups.google.com/forum/#!msg/votca/fGgjXpJQqmI/npxPH-2aPQAJ>

With Gromacs-2016 released on [Aug 5th 2016](https://mailman-1.sys.kth.se/pipermail/gromacs.org_gmx-users/2016-August/107718.html), please leave gromacs-5 alone as it is unsupported by upstream now.

schererc:
Ok, I should have probably known this.

So I will leave the script as it is and try with a more recent Gromacs version.

junghans:
Did it fix it?

junghans:
fixed?

schererc:
Yes, with the newest Gromacs version it worked with the -tableb option

Am 21.07.2017 16:40 schrieb "Christoph Junghans" <notifications@github.com>:

PR #219, opened 2017-08-13
enabling -multidir support for gmx (MPI)

abrukhno:
enabling -multidir option for gmx mdrun (MPI) via extra script run_gromacs_multidir.sh which is called from run_gromacs; it deals with all the additional input and proper restarts after errors (checked in IBI and IMC iterations); this necessitated slight amendements in other files (3, 5 in total)

The scripts allow to run mdrun (MPI) with '-multidir' option, which can be used to reduce the wall-time per iteration (IBI/IMC) while accumulating better stats than otherwise, due to different initial configurations and, hence, diverse trajectories. This is especially helpful to improve convergence of IMC. As a matter of fact, the IMC solver may take longer to run in between iterations than the MD simulation. Same setup with different starting points (but at the same temperature) can improve convergence of IBI as well. There is perhaps very little benefit for uniform and homogeneous systems. 

For reference in Gromacs documentation (2016.3), see here: http://manual.gromacs.org/documentation/2016.3/user-guide/mdrun-features.html

### DIFF
--- a/share/scripts/inverse/csg_table
+++ b/share/scripts/inverse/csg_table
@@ -96,6 +96,10 @@ postadd overwrite postadd_overwrite.sh
 postadd plot postadd_plot.sh
 postadd average postadd_average.sh
 
+# smoothing for updated potential to allow forgetting noise from previous iterations (rdfs)
+# the script is the same as postupd_smooth.sh
+postadd smooth postadd_smooth.sh
+
 #convergence checks
 convergence_check default convergence_check_default.sh
 
@@ -194,6 +198,7 @@ tables jackknife tables_jackknife.pl
 # interface to gromacs
 initstep gromacs initialize_step_genericsim.sh
 run gromacs run_gromacs.sh
+run multidir run_gromacs_multidir.sh
 clean gromacs clean_generic.sh
 presimulation gromacs run_gromacs.sh --pre
 pressure gromacs calc_pressure_gromacs.sh
--- a/share/scripts/inverse/functions_common.sh
+++ b/share/scripts/inverse/functions_common.sh
@@ -642,22 +642,22 @@ add_to_csgshare() { #added an directory to the csg internal search directories
   for dirlist in "$@"; do
     old_IFS="$IFS"
     IFS=":"
-    for dir in $dirlist; do
+    for dir in ${dirlist}; do
       #dir maybe contains $PWD or something
       eval dir="$dir"
-      [[ -d $dir ]] || die "${FUNCNAME[0]}: Could not find scriptdir $dir"
-      dir="$(globalize_dir "$dir")"
+      [[ -d ${dir} ]] || die "${FUNCNAME[0]}: Could not find scriptdir ${dir}"
+      dir="$(globalize_dir "${dir}")"
       if [[ $end = "yes" ]]; then
-        export CSGSHARE="${CSGSHARE}${CSGSHARE:+:}$dir"
-        export PERL5LIB="${PERL5LIB}${PERL5LIB:+:}$dir"
-        export PYTHONPATH="${PYTHONPATH}${PYTHONPATH:+:}$dir"
+        export CSGSHARE="${CSGSHARE}${CSGSHARE:+:}${dir}"
+        export PERL5LIB="${PERL5LIB}${PERL5LIB:+:}${dir}"
+        export PYTHONPATH="${PYTHONPATH}${PYTHONPATH:+:}${dir}"
       else
-        export CSGSHARE="$dir${CSGSHARE:+:}$CSGSHARE"
-        export PERL5LIB="$dir${PERL5LIB:+:}$PERL5LIB"
-        export PYTHONPATH="$dir${PYTHONPATH:+:}$PYTHONPATH"
+        export CSGSHARE="${dir}${CSGSHARE:+:}${CSGSHARE}"
+        export PERL5LIB="${dir}${PERL5LIB:+:}${PERL5LIB}"
+        export PYTHONPATH="${dir}${PYTHONPATH:+:}${PYTHONPATH}"
       fi
     done
-    IFS="$old_IFS"
+    IFS="${old_IFS}"
   done
   check_path_variable CSGSHARE PERL5LIB PYTHONPATH
 }
--- a/share/scripts/inverse/postadd_smooth.sh
+++ b/share/scripts/inverse/postadd_smooth.sh
@@ -1,0 +1,43 @@
+#! /bin/bash
+#
+# Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+if [ "$1" = "--help" ]; then
+cat <<EOF
+${0##*/}, version %version%
+This script implements smoothing of the updated potential (.pot)
+
+Usage: ${0##*/} infile outfile
+EOF
+   exit 0
+fi
+
+[[ -z $1 || -z $2 ]] && die "${0##*/}: Missing arguments"
+
+[ -f "$2" ] && die "${0##*/}: $2 is already there"
+
+name=$(csg_get_interaction_property name)
+tmpfile=$(critical mktemp ${name}.XXX)
+iterations=$(csg_get_interaction_property inverse.post_add_options.smooth.iterations)
+
+critical cp "$1" $tmpfile
+echo "doing $iterations smoothing iterations for interaction $name"
+
+for((i=0;i<$iterations;i++)); do
+  do_external table smooth $tmpfile "$2"
+  critical cp "$2" $tmpfile
+done
+
--- a/share/scripts/inverse/postadd_smooth.sh
+++ b/share/scripts/inverse/postadd_smooth.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 #
-# Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
+# Copyright 2009-2017 The VOTCA Development Team (http://www.votca.org)
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/share/scripts/inverse/run_gromacs.sh
+++ b/share/scripts/inverse/run_gromacs.sh
@@ -28,6 +28,27 @@ EOF
    exit 0
 fi
 
+tasks=$(get_number_tasks)
+
+mdrun_opts="$(csg_get_property --allow-empty cg.inverse.gromacs.mdrun.opts)"
+
+if [[ ${mdrun_opts} == *multidir* ]]; then
+
+  mdirs=$(echo ${mdrun_opts} | grep -o "sim" | wc -l)
+
+#check mdrun '-multidir' option for consistency
+
+  [[ -n ${mdirs} ]] || die "${0##*/}: mdrun '-multidir' option does not contain directory names including pattern 'sim' (check cg.inverse.gromacs.mdrun.opts)"
+
+  [[ ${mdirs} -gt 1 ]] || die "${0##*/}: mdrun '-multidir' option is set but the number of directories including pattern 'sim' <= 1 (check cg.inverse.gromacs.mdrun.opts)"
+
+  [[ ${mdirs} -le ${tasks} ]] || die "${0##*/}: mdrun '-multidir' option provided presumes the number of separate simulations greater than the number of tasks/threads (check cg.inverse.gromacs.mdrun.opts)"
+
+  do_external run multidir
+  exit
+
+fi
+
 tpr="$(csg_get_property cg.inverse.gromacs.topol)"
 
 mdp="$(csg_get_property cg.inverse.gromacs.mdp)"
@@ -37,7 +58,6 @@ conf="$(csg_get_property cg.inverse.gromacs.conf)"
 [[ -f $conf ]] || die "${0##*/}: gromacs initial configuration file '$conf' not found (make sure it is in cg.inverse.filelist)"
 
 confout="$(csg_get_property cg.inverse.gromacs.conf_out)"
-mdrun_opts="$(csg_get_property --allow-empty cg.inverse.gromacs.mdrun.opts)"
 
 index="$(csg_get_property cg.inverse.gromacs.index)"
 [[ -f $index ]] || die "${0##*/}: grompp index file '$index' not found (make sure it is in cg.inverse.filelist)"
@@ -133,14 +153,15 @@ else
   echo "${0##*/}: No walltime defined, so no time limitation given to $mdrun"
 fi
 
-#>gmx-5.1 has new handling of bonded tables
-if [[ ${mdrun_opts} != *tableb* ]]; then
+#gmx-5.1 has new handling of bonded tables - is it true only for ver.5.1 or above too???
+#if [[ ${mdrun_opts} != *tableb* ]]; then
+if [[ $(critical ${grompp} -h 2>&1) = *"VERSION 5.1"* && ${mdrun_opts} != *tableb* ]]; then
   tables=
   for i in table_[abd][0-9]*.xvg; do
     [[ -f $i ]] && tables+=" $i"
   done
   if [[ -n ${tables} ]]; then
-	  msg --color blue --to-stderr "Automatically added '-tableb${tables} to mdrun options (add -tableb option to cg.inverse.gromacs.mdrun.opts yourself if this is wrong)"
+    msg --color blue --to-stderr "${0##*/}: Automatically added '-tableb${tables} to mdrun options (add -tableb option to cg.inverse.gromacs.mdrun.opts yourself if this is wrong)"
     mdrun_opts+=" -tableb${tables}"
   fi
 fi
--- a/share/scripts/inverse/run_gromacs.sh
+++ b/share/scripts/inverse/run_gromacs.sh
@@ -154,7 +154,7 @@ else
 fi
 
 #gmx-5.1 has new handling of bonded tables
-    mdrun_opts
+
 if [[ ${mdrun_opts} != *tableb* ]]; then
   tables=
   for i in table_[abd][0-9]*.xvg; do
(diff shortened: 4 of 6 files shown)


share/scripts/inverse/csg_table
```diff
@@ -96,6 +96,10 @@ postadd overwrite postadd_overwrite.sh
 postadd plot postadd_plot.sh
 postadd average postadd_average.sh
 
+# smoothing for updated potential to allow forgetting noise from previous iterations (rdfs)
+# the script is the same as postupd_smooth.sh
+postadd smooth postadd_smooth.sh
```
junghans, line 101: Smoothing of the potential after adding (`postadd`) will lead to washing out of features, which is we don't have it, see https://groups.google.com/d/topic/votca/h2Lqi62Ncrs/discussion.

So, why does the normal post-update smoothing don't work for you?

share/scripts/inverse/functions_common.sh
```diff
@@ -642,22 +642,22 @@ add_to_csgshare() { #added an directory to the csg internal search directories
   for dirlist in "$@"; do
     old_IFS="$IFS"
     IFS=":"
```
junghans, line 644: All these changes do nothing? So why?

share/scripts/inverse/postadd_smooth.sh
```diff
@@ -0,0 +1,43 @@
+#! /bin/bash
+#
+# Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
```
junghans: 2017?

share/scripts/inverse/run_gromacs.sh
```diff
@@ -28,6 +28,27 @@ EOF
    exit 0
 fi
 
+tasks=$(get_number_tasks)
+
+mdrun_opts="$(csg_get_property --allow-empty cg.inverse.gromacs.mdrun.opts)"
+
+if [[ ${mdrun_opts} == *multidir* ]]; then
+
+  mdirs=$(echo ${mdrun_opts} | grep -o "sim" | wc -l)
```
junghans, line 37: Do the dir have to contain `sim`? Will an input file like `conf_sim.gro` break that?

share/scripts/inverse/run_gromacs.sh
```diff
@@ -28,6 +28,27 @@ EOF
...
+#check mdrun '-multidir' option for consistency
+
+  [[ -n ${mdirs} ]] || die "${0##*/}: mdrun '-multidir' option does not contain directory names including pattern 'sim' (check cg.inverse.gromacs.mdrun.opts)"
+
+  [[ ${mdirs} -gt 1 ]] || die "${0##*/}: mdrun '-multidir' option is set but the number of directories including pattern 'sim' <= 1 (check cg.inverse.gromacs.mdrun.opts)"
+
+  [[ ${mdirs} -le ${tasks} ]] || die "${0##*/}: mdrun '-multidir' option provided presumes the number of separate simulations greater than the number of tasks/threads (check cg.inverse.gromacs.mdrun.opts)"
+
+  do_external run multidir
+  exit
```
junghans: `exit 0` I guess.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -133,14 +153,15 @@ else
   echo "${0##*/}: No walltime defined, so no time limitation given to $mdrun"
 fi
 
-#>gmx-5.1 has new handling of bonded tables
-if [[ ${mdrun_opts} != *tableb* ]]; then
+#gmx-5.1 has new handling of bonded tables - is it true only for ver.5.1 or above too???
```
junghans: It is true for all versions >=5.1.3 and 5.1.[12] are broken with bonded tables anyhow, so undo.

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
+#! /bin/bash
+#
+# Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
```
junghans: 2017?

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+if [[ $1 = "--help" ]]; then
+cat <<EOF
+${0##*/}, version %version%
+This script runs a gromacs simulation or pre-simulation
```
junghans: Help needs update.

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
...
+[[ ${mdirs} -le ${tasks} ]] || die "${0##*/}: mdrun '-multidir' option provided presumes the number of separate simulations greater than the number of tasks/threads (check cg.inverse.gromacs.mdrun.opts)"
+
+echo -e "\nCreating/checking simulation data for ${mdirs} separate simulations (run in parallel)"
+
+#confout="confout.gro"
+confout="$(csg_get_property cg.inverse.gromacs.conf_out)"
+
+#conf="conf.gro"
+conf="$(csg_get_property cg.inverse.gromacs.conf)"
+if [[ -f ${conf}.${it} ]]; then
```
junghans: This is outside of the `mdirs`, so `$it` is always empty?

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
...
+    critical cp -f "${conf}.${it}" "$conf"
+fi
+[[ -f $conf ]] || die "${0##*/}: gromacs initial configuration file '$conf' not found (make sure it is in cg.inverse.filelist)"
+
+#mdp="grompp.mdp"
```
junghans: Remove comments, more time below

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
...
+	      [[ $(get_simulation_setting nstxtcout 0) -eq 0 ]] && die "${0##*/}: trajectory type (cg.inverse.gromacs.traj) is '${traj##*.}', but nstxtcout is 0 in $mdp. Please check the setting again and remove the current step."
+	   elif [[ ${traj} == *.trr ]]; then
+		[[ $(get_simulation_setting nstxout 0) -eq 0 ]] && die "${0##*/}: trajectory type (cg.inverse.gromacs.traj) is '${traj##*.}', but nstxout is 0 in $mdp. Please check the setting again and remove the current step."
+	   else
+	       die "${0##*/}: error trajectory type '${traj##*.}' (ending from '${traj}') is not supported"
+	   fi
+	fi
+
+        [[ -f ${checkpoint} ]] && echo "using checkpoint file '${checkpoint}'"
+
```
junghans, line 135: We need to find a better way, to not have to duplicate this piece of code.

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,380 @@
...
+#support for older mdp files (before ver.5.0): 
+#'cutoff-scheme = Verlet' is the default for Gromacs 5.0, but does not work with tabulated interactions
```
junghans, line 137: Gromacs-5.0 isn't support anymore, so fix comment.

junghans (changes requested):
Please provide an example in a separate pull-request for csg-tutorial.

share/scripts/inverse/csg_table
abrukhno: I did not mean to substitute smoothing the dU updates! This is just an extra option which I use at the beginning and intermediate stages of IBI or IMC: 2 or 3 smoothing loops for dU and 1 for U_new. I noticed that otherwise some strong noise features (due to poorly populated bins/regions) take very long to get smoothed (or can prevail unless I remove them manually).

share/scripts/inverse/functions_common.sh
abrukhno: Well, I was surprised myself, but without these changes (${var} vs £var) I was getting errors and could not progress with my testing. Sorry I included them now, in this commit, but otherwise it did not work for me.

share/scripts/inverse/run_gromacs.sh
abrukhno: Yep, 'sim' must be part of all the subdirectory names for the '-multidir' option in csg.inverse.gromacs.mdrun.opts. From the user perspective, it would be an error-prone overkill to have yet another option for the number of MPI-tasks =#subdirs < #tasks (overall threads), let alone an extra flag on top of these (I did consider <multidir>yes/no <number>#</number></multidir>). The way I did it the user only needs to specify (or not) the option itself and nothing else to support it. 

I don't get how having conf_sim.gro in the directory can break things down?

share/scripts/inverse/run_gromacs.sh
abrukhno: absolutely!

share/scripts/inverse/run_gromacs.sh
abrukhno: There is a bit of convention going on indeed, which must go into the manual eventually - the option has this structure: '-multidir sim_0 sim_1 sim_2 sim_3', i.e. the dir-names can be anything but should contain pattern 'sim' and end with sequential numbers starting with zero (the last requirement can be dropped, if really unwanted; upon some re-coding). 

All the input files must be provided with a suffix (ending) following the same convention: '.#' = dot followed by sequential numbers (starting with zero, not one). The master grompp which is used for the run collecting RDF stats, must have nsteps = mdirs * nsteps(subdir).

share/scripts/inverse/run_gromacs.sh
abrukhno: I really don't understand then why having it as you had originally could produce the error for me (gromacs ver. 5.1.2): too many input/table files provided (due to "-tableb ..." being explicitly added to the options by this piece of code). The error message is also confusing, as it suggests that I could fix it by providing the options via options.xml file, while I had already have that there, and deliberately(!) did not specify '-tableb ...' list.

abrukhno: But since you insist, I revoke this change. :)

share/scripts/inverse/run_gromacs_multidir.sh
abrukhno: When I did not have 'cutoff-scheme = Group' in my grompp files ver. 5.1.2 also did not run with tables and complained all the same, this is why I picked it up and changed "5.0" to just "5".

share/scripts/inverse/run_gromacs_multidir.sh
abrukhno: Sorry, I did not get what you mean here.

abrukhno (commented):
Attended most of the points raised. There is some controversy about Gromacs (ver. 5.*.*) behaviour, regarding 'cutoff-scheme = Group' vs 'Verlet' (default) and '-tableb ...' file-list option needed or not(?)

share/scripts/inverse/csg_table
junghans: I understand, this doesn't replace the smoothing the dU update, but I am wondering if we need this at all. Plus, it has nothing to do with the `multidir` feature, so let's do this in another pull request (and have the discussion there.)

(Background earlier version of VOTCA has a `postupd smooth`, but users always used it wrong and got a very washed out, feature-less, potentials.)

share/scripts/inverse/functions_common.sh
junghans: `$var` and `${var}` have identical meaning in bash! So, error message or it didn't happen!

share/scripts/inverse/run_gromacs.sh
junghans: My point was, if for some reason `sim` is part of another unrelated option in `${mdrun_opts}` this will break!

share/scripts/inverse/run_gromacs_multidir.sh
junghans: There is is a look to duplicated from `run_gromacs.sh` in your script and we should not do that....

share/scripts/inverse/run_gromacs.sh
junghans: The `-tableb` option is broken in gromacs-5.1.[012], see #210 for details and there is no workaround for that. Bonded tables just don't work in these versions!

And this block is just to ensure that if the user already has `-tableb` listed in `${mdrun_opts}`, we don't do anything.

On a related note, we should an error message about that bug.

share/scripts/inverse/run_gromacs_multidir.sh
junghans: Yes! I don't want to see gmx-5.0 anywhere in the code anymore.

junghans:
I moved the `tableb` issue into #220. And we should split out the add of `postadd_smooth.sh` into its own pull request.

For me, the main point of contention are:

- [ ] Code duplication with `run_gromacs.sh`
- [ ] Missing example in csg-tutorials

share/scripts/inverse/run_gromacs_multidir.sh
```diff
@@ -0,0 +1,342 @@
...
+	else
+        #get needed files 
+	    echo "Copying the needed files"
+	    [[ -n ${filelist} ]] && cp_from_main_dir "$filelist"
+
+	    mark_done "Filecopy"
+	fi
+	
+	if [[ -f ${conf}.${it} ]]; then
+    	   critical cp -f "${conf}.${it}" "$conf"
```
junghans, line 103: Another question, wouldn't all that copying around not be necessary if we use `-mulit` instead of `-multidir`?

share/scripts/inverse/run_gromacs.sh
junghans: Honestly, I would prefer an user option, instead of relying on a hard-coded pattern (`sim`) in the options.

I would do that in the same spirit as the `pre` simulation feature has its own top, gro, mdp files, but fails back to the main one if not specified.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -133,14 +153,15 @@ else
   echo "${0##*/}: No walltime defined, so no time limitation given to $mdrun"
 fi
 
-#>gmx-5.1 has new handling of bonded tables
+#gmx-5.1 has new handling of bonded tables
```
junghans, line 156: It is from >gmx-5.1, v2016 works the same way.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -133,14 +153,15 @@ else
...
+    mdrun_opts
```
junghans: A call to `mdrun_opts` ?

share/scripts/inverse/run_gromacs_multidir.sh
abrukhno: Meaning: you'd like to have a specific script for the currently repeated pieces? - in this case for the bits invoked by '-pre' option?

share/scripts/inverse/csg_table
abrukhno: Alright, how to branch/fork out this from here into a separate issue?

share/scripts/inverse/run_gromacs_multidir.sh
junghans: That or put all the extra stuff in `if [[ $multisim = yes ]];` block and unit your script with `run_gromacs.sh`.

share/scripts/inverse/csg_table
```diff
@@ -194,6 +198,7 @@ tables jackknife tables_jackknife.pl
 # interface to gromacs
 initstep gromacs initialize_step_genericsim.sh
 run gromacs run_gromacs.sh
+run multidir run_gromacs_multidir.sh
```
junghans, line 201: You could just do:
```
run multidir run_gromacs.sh --multi
```

share/scripts/inverse/run_gromacs.sh
```diff
@@ -133,14 +153,15 @@ else
...
+#gmx-5.1 has new handling of bonded tables
+
 if [[ ${mdrun_opts} != *tableb* ]]; then
   tables=
   for i in table_[abd][0-9]*.xvg; do
     [[ -f $i ]] && tables+=" $i"
   done
   if [[ -n ${tables} ]]; then
-	  msg --color blue --to-stderr "Automatically added '-tableb${tables} to mdrun options (add -tableb option to cg.inverse.gromacs.mdrun.opts yourself if this is wrong)"
+    msg --color blue --to-stderr "${0##*/}: Automatically added '-tableb${tables} to mdrun options (add -tableb option to cg.inverse.gromacs.mdrun.opts yourself if this is wrong)"
```
junghans: Honestly, I don't like to print the `${0##*/}` to the screen.

share/scripts/inverse/run_gromacs_multidir.sh
junghans: Or even `[[ $1 == --multi ]]`.

share/scripts/inverse/run_gromacs_multidir.sh
abrukhno: According to the latest documentation (http://manual.gromacs.org/documentation/2016.3/user-guide/mdrun-features.html), the main difference between -multidir and -multi is that in the former case separate simulation inputs/outputs are dealt with in separate directories vs in the latter case all files are dealt with from within the same directory but files recognised by the numeric suffix preceding the extension dot. 

I opted for the sub-directories (now I recall it!) as it is cleaner to check and I could abuse existing mechanisms in VOTCA for capturing errors and dealing with restarts etc (as all the names in a subdir are the same as for a single simulation. I admit, it is sort of non-elegant to copy and rename the initially suffixed input files but at least it is done in only one additional script (i.e. minimising my intervention into the existing VOTCA scripts).

With the '-multi' option everything would go in the same dir... run_gromacs.sh and possibly other gromacs-related scripts would need to be amended.

share/scripts/inverse/run_gromacs.sh
abrukhno: Guess why I do it this way? :) - what happens if I see an error or warning message and I want to know which script produced it?

share/scripts/inverse/csg_table
junghans: Just undo them here (on this branch) and then put them on a different branch (starting from `master`).

However, maybe we should discuss the use case for `postupd smooth` on the mailing list (or in an issue) first, because I still don't see the need for it.

@symashayak: any opinion on that?

share/scripts/inverse/run_gromacs.sh
abrukhno: I tried to follow the following rule: echo messages do not contain script names, critical warnings and errors do.

junghans: The is true for the logfile, but not for `msg`, which go the screen.  None of the other 91 `msg` has that.

share/scripts/inverse/csg_table
symashayak: Sorry for the late reply. I agree we should discuss this in a separate issue. 

Regarding the question of CG potential smoothing, I agree that due to CG updates may contain noise due to insufficient sampling. But, as per my understanding, I believe, it is more appropriate to address the noise in the update, i.e. dU, either by smoothing or (more accurately) by better sampling. As @junghans mentioned, smoothing the updated CG potential may inaccurately remove important features in the CG potential.

PR #221, opened 2017-08-15
run_gromacs.sh: warn user of gmx-5.1.[12]

### LINKED ISSUE #220: Add error message that bonded table don't work in gromacs-5.1.[012]
Bonded tables worked in <gromacs-5.0* and >=gromacs-5.1.3, but the version in between are broken. Usually that happens all the time and we would advice the user to simply update gromacs. However, ubuntu ships gromacs-5.1.2, which is on the red list, so a warning to the user would be nice.

/cc @abrukhno 

Refs: #210 and https://groups.google.com/forum/#!msg/votca/fGgjXpJQqmI/npxPH-2aPQAJ

junghans:
Fix #220


junghans:
Ok, that was supposed to fail on Ubuntu! And the error message got displayed correctly.

PR #222, opened 2017-08-23
REMD with gromacs

DRosen285:
I've added the possibility to use REMD,  instead of normal MD when using gromacs.


junghans:
Could you do a merge with the latest master please?

junghans:
Done!

junghans (changes requested):
Minor nitpick:
- [ ] Indentation
- [ ] Tags in `csg_defaults.xml` should be lower case

Major:
- [ ] `mdrun` should at least honor `CSG_MDRUN_STEPS` and use `gromacs_log`
- [ ] Example in csg-tutorials

junghans:
One other thought, instead of:
```
if [[ $(csg_get_property cg.inverse.gromacs.REMD) = "yes" ]]; then
  bla
fi
```
we should do
```
if [[ $(csg_get_property cg.inverse.gromacs.runmode) = "replex" ]]; then
  bla
fi
```
and that will make integration with #219 easier. The default of `runmode` should be `none`.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -115,12 +117,32 @@ if [[ ${CSG_MDRUN_OPTS} ]]; then
...
+if [[ $(csg_get_property cg.inverse.gromacs.REMD) = "yes" ]]; then
+PREFIX="$(csg_get_property cg.inverse.gromacs.REMD.PREFIX)"
+read -a T <<<"$(csg_get_property cg.inverse.gromacs.REMD.T_for_replicas)"
+j=0
+for i in "${T[@]}"
+do
+echo $i
+#TEMP has to be written as a place holer for the temperature in the *.mdp file
+sed "s/TEMP/$i/" grompp.mdp > grompp_$j.mdp
+critical ${grompp[@]} -n "${index}" -f "grompp_$j.mdp" -p "$topol_in" -o "$PREFIX"_"$j.tpr" -c "${conf}" ${grompp_opts} 2>&1 | gromacs_log "${grompp[@]} -n "${index}" -f "${mdp}" -p "$topol_in" -o "$tpr" -c "${conf}" ${grompp_opts}"
```
junghans: Would it be too much effort for the user to make `grompp_1.mdp` ... `grgrompp_n.mdp` by hand and list them in `filelist` explicitly? (And the same for `conf*.gro`?)

@DRosen285 , @abrukhno : Opinions?

DRosen285: I thought it's more user friendly , if they just have to change one line in the xml file, instead of changing one line in n files (related to the mdp files). But if it's better to parse the files via the list, it's possible to do so

abrukhno: Hi both, I am trying to follow what is done here: 

1) replica-exchange 'REMD' option is searched in options.xml by the tag 'cg.inverse.gromacs.REMD' and, if it's found with value "yes", then an REMD branch in run_gromacs.sh is followed (this is clear, just for the sake of completeness)

2) some PREFIX is read in from the tag 'cg.inverse.gromacs.REMD.PREFIX' - what is this prefix? - It looks like "$PREFIX" plus the replica index is the .tpr file name; Why calling the tag 'PREFIX' then? - would it be clearer (self-explanatory for the user) to call it what it is: 'REMD.tpr_name'? It does not seem that PREFIX is used as anything else(?). As it is, at first look I was anticipating it to be a prefix for some standard file name(s).

3) A set of Temperatures is read in from options.xml tag 'cg.inverse.gromacs.REMD.T_for_replicas' - to my liking 'T_for_replicas' is too long for a tag, e.g. I would prefer 'replicas_T', but that's a matter of taste of course

4) Using template grompp.mdp file: each T value is added into it by replacing the place-holder "TEMP" (if it's found in grompp.mdp) and saving under the name grompp_$j.mdp (suffix being the replica index). 

What I dislike: grompp.mdp needs to have a place-holder TEMP to be replaced with an actual option, i.e. this grompp cannot be used by itself, without modification. Even though it is just one line, the user will have to remember and take care of that line. - as an alternative, I would replace the existing(!) option 'ref-t ='*  in grompp.mdp:

sed '/ref-t/s/.*/$i/' grompp.mdp > grompp_$j.mdp

Another thin is that "grompp" is hard-coded as the file name, whereas the user might what to have some generic name-pattern for all the files: my_name.mdp, my_name.top, my_name.gro etc.

Sorry for this being so long / Andrey

abrukhno (commented):
Please, see my lengthy comment. 
To summarise, we need to come up with unified conventions for tag naming and scripting both REMD #222 and MULTI #219 (assuming they would be different options, which I am not sure they should be) - ?

share/scripts/inverse/run_gromacs.sh
DRosen285: Hi Andrey, thanks for the comment. Let me quickly reply:

To 2.: Prefix is also used for the output trajectories. And I basically followed the explanations and syntax of gromacs. (see http://www.gromacs.org/Documentation/How-tos/REMD)

To 3: I like your suggestion

To 4: Can you specify what do you mean by replacing 'ref-t' ? Do you mean to leave the right hand of 'ref t' blank or deleting the whole line in the .mdp file and parse it within the run_gromacs script ? 

Cheers,
David

junghans:
@abrukhno: I was thinking of something like:
```bash
if [[ $(csg_get_property cg.inverse.gromacs.runmode) = "replex" ]]; then
  bla
elseif [[ $(csg_get_property cg.inverse.gromacs.runmode) = "multi" ]]; then
  blu
else
  foo
fi
# common code
```

share/scripts/inverse/run_gromacs.sh
abrukhno: Hi David, thanks for your insights. The link does not describe the workflow completely, so I can only assume that:

(a) all the input files must be repeated N_replicas times, their names taking the form <file-name>_<replica-index>.<extension>, as it is for '-multi' option (which is anyway/always needed for REMD, apparently); 
NOTE that, according to another description: http://manual.gromacs.org/documentation/2016.3/user-guide/mdrun-features.html , one can use '-multidir' instead of '-multi', in which case the user has to prepare N_replicas (indexed) subdirectories that should contain input files for each replica separately, and no indexing in file names is needed, i.e. one can use grompp.mdp, config.mdp topol.top etc which may contain different inputs.

(b) correspondingly, the output will comprise N_replicas of output files named according to the chosen option, either '-multi' or '-multidir', in the latter case the output files for each replica are stored in the corresponding (indexed) subdirectories. 

(c) post-simulation analysis should either extract/combine the trajectories per ensemble (a single trajectory per T value) or per replicas time (a single trajectory per replica, where T is varying), or both, depending on the user's task/requirements.

Based on the above, I suggest that we do not split this workflow into REMD/multi branches (as suggested by different 'runmode's), but rather combine the two within the same branch. The tags might work like this: 'cg.inverse.gromacs.replicas'=number, >0 for '-multi'/'-multidir' options (default=0, meaning normal mdrun run), and 'cg.inverse.gromacs.replicas.ref-ts'=string (default ="", i.e. empty), which for REMD won't be empty and contain a set of T-values. 

Then the user will have full control over the natural Gromacs workflows and have freedom to decide on the bookkeeping wrt input and output files etc. We will have an extra complication, though, to arrange the concatenation of trajectory file according to another option: 'cg.inverse.gromacs.replicas.trjcat_opts'

What do you think? - Christoph, you are welcome to comment too!

abrukhno: sorry, the internal parsing here ate my input for file names: file-name_index.extension (I tried to use angle brackets above, for no avail).

abrukhno (commented):
Since REMD is a subset (or equivalent) with respect to '-multi' and '-multidir' options, I suggest to combine the two 'runmodes' (i.e. have just one). 

The needed tags might be the following: 'cg.inverse.gromacs.replicas'=number, >0 for '-multi'/'-multidir' options (default=0, meaning normal mdrun run), and 'cg.inverse.gromacs.replicas.ref-ts'=string (default ="", i.e. empty), which for REMD won't be empty and contain a set of T-values.

The mdrun options should be checked for '-multi'/'multidir' and '-repex' options, and extra tag would be needed for trjcat options:
'cg.inverse.gromacs.replicas.trjcat_opts'

It will of course require quite some re-coding within either my #219 or David's #222 branches.

junghans:
I started working on this, still work in progress, see #223

junghans (approved):
See #223

PR #223, opened 2017-08-30
Add replica dynamics (and multi) support

junghans:
Fix #222 
Fix #219 

Tutorials here: votca/csg-tutorials#8


junghans:
@DRosen285, @abrukhno 

1. Can you think of a case, where we don't want to do a concatenation of the trajectories automatically for `-multi` (but without `-replex`).
2. For `-replex` should we always use the 0th trajectories? Maybe that isn't the temperature of interest or we are doing Hamiltonian Replica Exchange! We could add an option to specify, which trajectories gets handed to `csg_stat`, `csg_stat --do-imc` and `csg_re`, similar to the overwrite of the topology files for `csg_stat`.
3. Currently this won't work for `cg.inverse.initial_configuration` being `laststep`, should we support that? I think this won't make much sense for `-regex`, but maybe for `-multi`?

junghans:
Ok, this is work now, so comments are welcome.

abrukhno:
Hi Christoph,

You keep ahead of us! I am replying below:

>1.    Can you think of a case, where we don't want to do a concatenation of the trajectories automatically for -multi (but without -regex).

What is '-regex'? - could not find it in GMX manual, so I read it as '-replex' for REMD. Then different scenaria of pulls and lambda sampling might be the cases where one would only use -multi but not -replex.

>2.    For -regex should we always use the 0th trajectories? Maybe that isn't the temperature of interest or we are doing Hamiltonian Replica Exchange! We could add an option to specify, which trajectories gets handed to csg_stat, csg_stat --do-imc and csg_re, similar to the overwrite of the topology files for csg_stat.

This is a tricky question: do you anticipate reweighting (WHAM) to other T than the min value? normally one wants to Boltzmann-project the stats to the lowest T value, but in the past I usually used all stats projected to all T values, so it is debatable. It also depends - does the 0-th trajectory always correspond to the lowest T?

>3.    Currently this won't work for cg.inverse.initial_configuration being laststep, should we support that? I think this won't make much sense for -regex, but maybe for -multi?

I have no strong feeling for this option, as I never used laststep for the initial setup of the next iteration.

Cheers / Andrey

abrukhno:
...forgot: regarding point 2, I foresee (anticipate) that rerunning / replaying trajectory for recollection of stats might be a good option for histogram reweighting. I don't know however if WHAM utility supports that(?).

DRosen285:
Hey,

sorry for being a little late. But I was a bit struggling in fully understanding the code.
Just for clarification: If the user defines -multi or -regex in the mdrun_opts then the different commands from grompp etc. are called ?
And is there a way to use -regex without -multi ? I don't see that, but probably that's also not necessary.
 
Anyways my comments 
> 1. Can you think of a case, where we don't want to do a concatenation of the trajectories automatically for -multi (but without -replex).

I've never really used multidir, just read about it. So I can't think of a case.

> 2. For -regex should we always use the 0th trajectories? Maybe that isn't the temperature of interest or we are doing Hamiltonian Replica Exchange! We could add an option to specify, which trajectories gets handed to csg_stat, csg_stat --do-imc and csg_re, similar to the overwrite of the topology files for csg_stat.

No we should not always use the 0th trajectories. So I think an additional option for specification would be nice.

> 3. Currently this won't work for cg.inverse.initial_configuration being laststep, should we support that? I think this won't make much sense for -replex, but maybe for -multi?

Regarding this point I'm going d'accord with Andrey.

Cheers,
David

junghans:
>> Can you think of a case, where we don't want to do a concatenation of the trajectories automatically for -multi (but without -regex).

> What is '-regex'? - could not find it in GMX manual, so I read it as '-replex' for REMD. Then different scenaria of pulls and lambda sampling might be the cases where one would only use -multi but not -replex.

I meant `-replex` of course.

> This is a tricky question: do you anticipate reweighting (WHAM) to other T than the min value? normally one wants to Boltzmann-project the stats to the lowest T value, but in the past I usually used all stats projected to all T values, so it is debatable. It also depends - does the 0-th trajectory always correspond to the lowest T?

No, I don't want to support in the main code for now, there is a `pre_update` script the user could overwrite to do exactly that. I think, with `-replex` one could do a simulation with different Hamiltonians and then the 0-th trajectory might be some kind of Go-like ground state model, which you most certain don't want to calculate the data from.

For 3.) I will have a look how hard it will be to implement.

junghans:
> No we should not always use the 0th trajectories. So I think an additional option for specification would be nice.

Ok, I was about to add an option for that anyhow.

junghans:
Ok, this is now ready for a final review. I didn't add support for `laststep`, but just added an error message - we can add that feature later if somebody asks.

I also update votca/csg-tutorials#8 accordingly.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -30,31 +30,42 @@ fi
...
 grompp=( $(csg_get_property cg.inverse.gromacs.grompp.bin) )
 [[ -n "$(type -p ${grompp[0]})" ]] || die "${0##*/}: grompp binary '${grompp[0]}' not found"
 
 traj=$(csg_get_property cg.inverse.gromacs.traj)
-if [[ $1 != "--pre" ]]; then
-  #in a presimulation usually do care about traj and temperature
+# in main simulation we usually do care about traj and temperature
+# in pre-simulation/multisim there might be no traj and a different temp(s) 
+if [[ ! ${multi} && $1 != "--pre" ]]; then
+  # no temp check for multi simulation
```
junghans: Remove this line!

share/scripts/inverse/run_gromacs.sh
```diff
@@ -150,5 +184,29 @@ fi
...
+    [[ ! -f ${confout_x} ]] || [[ -z "$(sed -n '/[nN][aA][nN]/p' ${confout_x})" ]] || die "${0##*/}: There is a nan in '${confout_x}', this seems to be wrong."
+  done
+else
+  [[ ! -f ${confout} ]] || [[ -z "$(sed -n '/[nN][aA][nN]/p' ${confout})" ]] || die "${0##*/}: There is a nan in '${confout}', this seems to be wrong."
+fi
+
+if [[ ${multi} ]]; then
+  echo "Dummy file created by ${0##*/}" > "${confout}" #for simulation_finish(), confout isn't really used.
+  if [[ ${mdrun_opts} = *-replex* ]]; then
+     : #user has to tell imc, rdf, re updater what to use.
```
junghans: remove `:`.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -150,5 +184,29 @@ fi
...
+     echo "Dummy file created by ${0##*/}" > "${traj}" # for simulation_finish
```
junghans: `[[ -f ${traj} ]] || ...`

share/scripts/inverse/run_gromacs.sh
```diff
@@ -150,5 +184,29 @@ fi
...
+  for((i=0;i<${multi};i++)); do
+    confout_x="${confout%.*}${i}.${confout##*.}"
+    [[ ! -f ${confout_x} ]] || [[ -z "$(sed -n '/[nN][aA][nN]/p' ${confout_x})" ]] || die "${0##*/}: There is a nan in '${confout_x}', this seems to be wrong."
+  done
+else
+  [[ ! -f ${confout} ]] || [[ -z "$(sed -n '/[nN][aA][nN]/p' ${confout})" ]] || die "${0##*/}: There is a nan in '${confout}', this seems to be wrong."
+fi
+
+if [[ ${multi} ]]; then
+  echo "Dummy file created by ${0##*/}" > "${confout}" #for simulation_finish(), confout isn't really used.
```
junghans: `[[ -f ${confout} ]] || ...`

junghans:
@abrukhno @DRosen285 please review

share/scripts/inverse/run_gromacs.sh
```diff
@@ -115,12 +119,42 @@ if [[ ${CSG_MDRUN_OPTS} ]]; then
...
 fi
 
-#see can run grompp again as checksum of tpr does not appear in the checkpoint
-critical ${grompp[@]} -n "${index}" -f "${mdp}" -p "$topol_in" -o "$tpr" -c "${conf}" ${grompp_opts} 2>&1 | gromacs_log "${grompp[@]} -n "${index}" -f "${mdp}" -p "$topol_in" -o "$tpr" -c "${conf}" ${grompp_opts}"
-[[ -f $tpr ]] || die "${0##*/}: gromacs tpr file '$tpr' not found after runing grompp"
+# run gromacs multiple or one time
+if [[ ${multi} ]]; then
+  for((i=0;i<${multi};i++)); do
+    for j in index mdp topol_in conf; do
+      # read $j.$i (e.g index.1} and use ${!j} (e.g. ${index} ) as default and store it in f
```
DRosen285: Is it really $j.$i ?  Or should it be $j.$i.dat; I was thinking on sth. like index.1.ndx ?
I anyway have problems in running this script. It says It cannot find the mdp.files, although I've specified them under filelist. So is this a bug or a problem related to the .xml file ?

junghans: I think the comment is misleading, it should be `cg.inverse.gromacs.${j}.sim${i}`, where `$j` is the word `index`, `mdp` etc and `$i` is 0..n, so e.g. `cg.inverse.gromacs.mdp.sim2` is read.

share/scripts/inverse/run_gromacs.sh
```diff
@@ -115,12 +119,42 @@ if [[ ${CSG_MDRUN_OPTS} ]]; then
...
+      f="$(csg_get_property --allow-empty "cg.inverse.gromacs.${j}.sim${i}" "${!j}")" #filter me away
+      [[ -f ${f} ]] || die "${0##*/}: file '$f' not found (make sure it is in cg.inverse.filelist)"
+      read ${j}_x <<< "${f}" # set ${j}_x (topol_x to ${f}
```
junghans: missing "e.g.' here

share/scripts/inverse/run_gromacs.sh
junghans: And @DRosen285, do that work https://github.com/votca/csg-tutorials/blob/dc911a7a828a8541d1c96d4d38eb1590f0609c58/methanol/imc_gromacs-remd/settings.xml#L63 ?

DRosen285: Great. Now it's clear and everything is working :)

junghans:
FYI here is how you approve a pull request: https://help.github.com/articles/approving-a-pull-request-with-required-reviews/

share/scripts/inverse/run_gromacs.sh
abrukhno: Hi, in the following piece I do not understand - why the file with suffix '1' is missing, i.e. no 'sim1'?
`      <mdp>
        <sim0>grompp0.mdp</sim0>
        <sim2>grompp2.mdp</sim2>
        <sim3>grompp3.mdp</sim3>
      </mdp>
`

abrukhno (approved):
Hi, Sorry for the delay. 
After David's review I could not spot any issues now. Considering that the checks and the spce tutorial have passed ok, I am approving now, not to hold it up (although I did not have time yet to thoroughly test it).
Many thanks for adding this extension!

share/scripts/inverse/run_gromacs.sh
junghans: You could add `<sim1>grompp.mdp</sim1>`, but `grompp.mdp` is the default anyhow.

junghans: Plus, you would need to set the normal `<mdp>` option to tell VOTCA where look up default settings in `get_simulation_setting()`.

So:
```
    <mdp>grompp1.mdp
        <sim0>grompp0.mdp</sim0>
        <sim1>grompp1.mdp</sim1>
        <sim2>grompp2.mdp</sim2>
        <sim3>grompp3.mdp</sim3>
      </mdp>
```

DRosen285 (approved):

PR #228, opened 2017-09-20
Allow to override build date

bmwiedemann:
that ends up in man page headers
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This date call only works with GNU date.

An alternative approach that also works with BSD date is to use
`date -u -r CHANGELOG.md`

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -1,3 +1,4 @@
+string(TIMESTAMP VOTCA_DATE "%Y-%m-%d %H:%M:%S" UTC)
 configure_file(help2t2t.in ${CMAKE_CURRENT_BINARY_DIR}/help2t2t.out @ONLY)
 add_custom_target(help2t2t_build DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/help2t2t)
 add_custom_command(OUTPUT help2t2t ALL COMMAND ${CMAKE_COMMAND}
--- a/scripts/help2t2t.in
+++ b/scripts/help2t2t.in
@@ -72,7 +72,7 @@ fi
 cat <<EOF
 $prog
 Version: $version
-$(date)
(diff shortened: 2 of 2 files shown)


junghans:
What about `version.cc`? Is uses `__DATE__`, that isn't reproducible either.

bmwiedemann:
`__DATE__`  is reproducible, because gcc has patches to use $SOURCE_DATE_EPOCH for that => https://gcc.gnu.org/ml/gcc-patches/2016-11/msg00183.html

junghans:
@bmwiedemann: cool.

I honestly think `$(date -u -d @${SOURCE_DATE_EPOCH:-$(date +%s)})` is a bit convoluted. 

As we are pre-processing `help2t2t.in` with cmake's `configure_file()` I would like to catch SOURCE_DATE_EPOCH in cmake, something like:
```cmake
if (ENV{'SOURCE_DATE_EPOCH}')
  set(VOTCA_DATE ${ENV{'SOURCE_DATE_EPOCH}')
else()
  set(VOTCA_DATE "$(date)")
endif()
```
and replace `$(date)` by `@VOTCA_DATE` in `help2t2t.in`.

bmwiedemann:
with cmake 2.8.11 and later, you can just use
```
 string(TIMESTAMP VOTCA_DATE "%Y-%m-%d %H:%M:%S" UTC)
```
and latest cmake already has a patch to use `SOURCE_DATE_EPOCH` for it, which makes it a very nice solution from the reproducible-builds perspective.

bmwiedemann:
updated my PR to implement that nicer, less convoluted approach.
Also this is using UTC timezone and ISO date format (the only true way to present date/time)

junghans (approved):
Nice!

PR #230, opened 2017-11-24
#229 Updating pdbreader 

### LINKED ISSUE #229: PDB Reader Update
Apparently pdb reader does not have the ability to read the connectivity information from .pdb files. This needs to be added.

JoshuaSBrown:
Fix #229 
pdbreader has been split into two separate files one for reading the topological information and the second for reading the trajectories. 
The topological reader looks for the key word CONECT to determine which atoms are part of a given molecule, thus conserving the topological information. 
The trajectory reader only considers the atom .xyz positions.


include/votca/csg/atomtable.h
```diff
@@ -0,0 +1,137 @@
...
+
+    protected:
+
+        /*
+         * Return the index associated with the tables for the chosen element.
+         */
+        int getIndex(std::string sym);
+
+        /* 
+         * Stores the symbols for all the atoms in the periodic table.
```
jkrajniak: Instead of having vector of symbols, vector of masses and vector of atomic number, why not to have an unordered_map with the key being the symbol and the value being a struct with mass, atomic number
```c++
struct Atom {
  double mass;
  double atomicNumber;
};

std::unordered_map<std::string, Atom> atomMap;
atomMap.insert(std::make_pair("Ac", Atom{227.028, 89}));
...
```

You can of course in the constructor reuse the vectors that you already created and fill up the map in some loop.

With this you can reduce the complexity of all of the getters that you have.

src/libcsg/modules/io/pdbtopologyreader.cc
```diff
@@ -0,0 +1,415 @@
+/*
+ * Copyright 2009-2013 The VOTCA Development Team (http://www.votca.org)
```
jkrajniak: this should be 2017 or 2009-2017

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,82 @@
+/* 
+ * Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
```
jkrajniak: this should be 2009-2017 or 2017

include/votca/csg/atomtable.h
```diff
@@ -0,0 +1,137 @@
+/* 
+ * Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
```
jkrajniak: 2017 ?

src/libcsg/modules/io/pdbtopologyreader.cc
```diff
@@ -0,0 +1,415 @@
...
+                stringstream ss;
+                ss << string(line.substr(6));
+                // 1 -  6       Record name    "CONECT"
+                //11 -  7       Real(5)        atm1           (ID)
+                ss >> atm1; 
+                string temp_atm;          
+                ss >> temp_atm;
+                bonded_atms.push_back(temp_atm);
+                num_bonds = 1;
+                cout << atm1 << " " << temp_atm << " ";
```
jkrajniak: is it for debug? why do you output something here?

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,82 @@
...
+    int ind = this->getIndex(sym);
+    if(ind>=this->Atoms.size()){
+        cerr << "Error Unrecognized symbol in atomtable.cc " << sym << endl;
+        exit(1);
+    } 
+
+    double mass=this->MassNumber.at(ind);
+    if(mass==-1.0){
+        cerr << "Error mass for " << sym << " is unknown" << endl;
+        exit(1);
```
jkrajniak: this should throw an exception and not exit

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,82 @@
...
+vector<string> AtomTable::getNoble(void){
+  vector<string> noble = {"He","Ne","Ar","Kr","Xe","Rn","Uuo"};
+  return noble;
+}
+
+double AtomTable::getMass(string sym){
+    int ind = this->getIndex(sym);
+    if(ind>=this->Atoms.size()){
+        cerr << "Error Unrecognized symbol in atomtable.cc " << sym << endl;
+        exit(1);
```
jkrajniak: throw exception

jkrajniak (changes requested):

src/libcsg/modules/io/pdbtopologyreader.cc
JoshuaSBrown: Ooops forgot to remove all the cout statements.

junghans (changes requested):
Why do we need to split it into two files? Just put the topology-specific part in `if(_topology)` (see `groreader.cc`), this avoids code duplication and keep pdb reader code in one file.

JoshuaSBrown:
@junghans fine by me I was just trying to follow what other people had done. Is there a particular reason the gmx, dlpoly and mdp file types are split up. I can go ahead and combine those as well while I'm working on this if it will lead to more consistency.

junghans:
Well, we are trying to avoid code duplication! And all the readers work by file extension, so for gmx tpr is the topology extension and trr/xtc the trajectory extension.

Map is obsolete and not even compiled-in anymore!

dlpoly has two vastly different files for topology and trajectory!

I think pdb is most similar to gro, as it can serve as topology as well as trajectory format.

src/libcsg/modules/io/pdbreader.h
```diff
@@ -32,34 +33,39 @@ using namespace std;
...
     private:
+        string _fname;
         ifstream _fl;
         bool _topology;
 };
 
 }}
 
-#endif // __VOTCA_CSG_PDBREADER_H
+#endif // __VOTCA_CSG_PDBTOPOLOGYREADER_H
```
junghans: Huh?

src/libcsg/topologyreader.cc
```diff
@@ -1,5 +1,5 @@
 /* 
- * Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
+ * Copyright 2009-2017 The VOTCA Development Team (http://www.votca.org)
```
junghans: If you don't touch the code, don't change it!

src/libcsg/trajectoryreader.cc
```diff
@@ -1,5 +1,5 @@
 /* 
- * Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
+ * Copyright 2009-2017 The VOTCA Development Team (http://www.votca.org)
```
junghans: Same!

JoshuaSBrown:
Just a heads up I don't have a ton of experience using git. So if you see something that could be clearer or done better let me know.

src/libcsg/modules/io/pdbreader.cc
```diff
@@ -27,25 +29,35 @@ using namespace std;
...
+    boost::filesystem::path filepath(file.c_str());
+
+    // Check that the file has the correct extension
+    if(boost::filesystem::extension(filepath).size()==0 ){
+        throw std::ios_base::failure("ERROR on opening .pdb file'"
+                +file+"' - extension is expected, use .pdb");
+    }else if(boost::filesystem::extension(filepath)!=".pdb"){
+        throw std::ios_base::failure("Error on opening .pdb file '"
+                +file+"' - wrong extension, use .pdb");
+    }
```
junghans: Why is that needed? The trj/topo reader factory already make sure that this code is only called for a pdb files. Also if it wouldn't be called pdb (but e.g. pdbx) that would be fine, too.

src/libcsg/modules/io/pdbreader.cc
```diff
@@ -57,128 +69,383 @@ void PDBReader::Close()
...
+            double aa = boost::lexical_cast<double>(a)/10.0;
+            double bb = boost::lexical_cast<double>(b)/10.0;
+            double cc = boost::lexical_cast<double>(c)/10.0;
+	        top.setBox(matrix(vec(aa, 0 , 0 ),
+	                          vec(0 , bb, 0 ),
+                              vec(0 , 0 , cc)));
+
+	    }
+        // Only read the CONECT keyword if the topology is set too true
+        if( _topology && wildcmp("CONECT*",line.c_str())){
```
junghans, line 117: What happens if there is a `CONECT` line in a pdb trajectory?

src/libcsg/topologyreader.cc
junghans: Undo.

include/votca/csg/atomtable.h
junghans: I like the idea of an unordered_map.

src/libcsg/modules/io/pdbreader.cc
JoshuaSBrown: It was in one of the other file readers which I used as a template. I can remove it.

src/libcsg/modules/io/pdbreader.cc
JoshuaSBrown: As the code is currently written it will be ignored. That is if I remember correctly. _topology is set to true only when ReadTopology is called.

src/libcsg/modules/io/pdbreader.cc
junghans: The dlpolyreader can handle two extensions, plus it actually opens files with hard-codes names depending on that extension (yes, that is how dlpoly works!), so don't worry about that for the pdbreader and drop that block!

src/libcsg/modules/io/pdbreader.cc
junghans: Ok, i see it now. The while loop just ignores all lines with unknown tags! That is good!

junghans:
Btw, Travis is failing due to these warnings:
```
/home/votca/votca/csg/src/libcsg/atomtable.cc:32:22: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     for(int ind=0;ind<this->Atoms.size();ind++){
                   ~~~^~~~~~~~~~~~~~~~~~~
/home/votca/votca/csg/src/libcsg/atomtable.cc: In member function 'int votca::csg::AtomTable::getIndex(std::__cxx11::string)':
/home/votca/votca/csg/src/libcsg/atomtable.cc:48:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
```

src/libcsg/topologyreader.cc
JoshuaSBrown: Does it matter if I just change it back as I have. Or do you want me to go back to the commit where this was made and remove it from the commit history.

junghans: Just undo it for now, we can later decide it want to squash all commits of this PR into one depending on how much it is in total!

include/votca/csg/atomtable.h
```diff
@@ -0,0 +1,151 @@
...
+        /**
+         * Constructor
+         */
+        AtomTable(void);
+        /* 
+         * Destructor 
+         */
+        ~AtomTable() {}
+
+        /* *
```
junghans: `/**` without space I guess.

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,90 @@
...
+        string err_msg = "Invalid atom symbol in AtomTable::getMass. There is"
+                         " no atom with symbol "+sym;
+        throw invalid_argument(err_msg);
+    }
+    return mass;
+}
+
+int AtomTable::getAtomicNumber(string sym){
+    double num = atomMap[sym].atomicNumber;
+    if(num==0.0){
```
junghans: Double comparison against 0.0? That go wrong either use num < 1. And shouldn't atomic be an integer?

src/libcsg/modules/io/pdbreader.cc
```diff
@@ -27,25 +29,32 @@ using namespace std;
...
-{
+bool PDBReader::Open(const string &file){
+
+    boost::filesystem::path filepath(file.c_str());
+
+    // Check that the file has the correct extension
+    if(boost::filesystem::extension(filepath).size()==0 ){
+        throw std::ios_base::failure("ERROR on opening .pdb file'"
+                +file+"' - extension is expected, use .pdb");
+    }
```
junghans: I think that is not needed, as `TrjReaderFactory()` already checks the extension and only call the pdb reader if it is a `pdb` file.

src/libcsg/modules/io/pdbreader.h
```diff
@@ -23,39 +23,45 @@
...
-        /// open a trejectory file
+        /// open a trajectory file
         bool Open(const string &file);
         /// read in the first frame
         bool FirstFrame(Topology &top);
         /// read in the next frame
         bool NextFrame(Topology &top);
+        /// set the topology file name: <name>.pdb (convention: ".pdb"="FIELD")
+        void setFname(string name) { _fname = name; return; }
+        /// get the topology file name: <name>.dlpf (convention: ".pdb"="FIELD")
```
junghans: `.dlpf`?

junghans (changes requested):

src/libcsg/modules/io/pdbreader.h
JoshuaSBrown: Hmm, not sure how that got in there. You know I didn't realize I had such a habit of introducing typos like this until now.

src/libcsg/modules/io/pdbreader.cc
junghans: Drop this!

junghans (changes requested):
And I would still do https://github.com/votca/csg/pull/230#discussion_r153022674

src/libcsg/modules/io/pdbreader.h
```diff
@@ -23,39 +23,45 @@
...
         bool ReadTopology(string file, Topology &top);
-
-        /// open a trejectory file
+        /// open a trajectory file
         bool Open(const string &file);
         /// read in the first frame
         bool FirstFrame(Topology &top);
         /// read in the next frame
         bool NextFrame(Topology &top);
+        /// set the topology file name: <name>.pdb (convention: ".pdb"="FIELD")
```
junghans: Hmm, I think the `convention` part is wrong here and 2 lines below!

junghans:
@MrTheodor: please also have a 2nd pass!

src/libcsg/modules/io/pdbreader.h
JoshuaSBrown: What should it be?

junghans: No convention!

JoshuaSBrown: Ah, ok, I wasn't sure what the comment meant i the first case. I'll fix it.

junghans: Wait for @MrTheodor's comments too.

JoshuaSBrown: I'm guessing this is true for both the set and get functions?

junghans: Yes!

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,91 @@
...
+
+    for(unsigned int ind=0;ind<this->Atoms.size();ind++){   
+        Atom atm;
+        atm.mass         = MassNumber.at(ind);
+        atm.atomicNumber = AtomicNumber.at(ind);
+        atomMap.insert(std::make_pair(Atoms.at(ind),atm));
+    }
+}
+
+int AtomTable::getIndex(string sym){
```
jkrajniak: Do you still need this method?

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,91 @@
...
+        if(*it==sym){
+            return ind;
+        }
+    }
+    return ind;
+}
+
+/* Check if a string is a legitimate atomic symbol */
+bool AtomTable::checkSymbol(string sym){
+  for(auto it=this->Atoms.begin();it!=this->Atoms.end();it++){
```
jkrajniak: You already have atomMap with symbols, you can do the checkup by
```c++
return atomMap.find(sym) != atomMap.end();
```

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,91 @@
...
+}
+
+vector<string> AtomTable::getNoble(void){
+  vector<string> noble = {"He","Ne","Ar","Kr","Xe","Rn","Uuo"};
+  return noble;
+}
+
+double AtomTable::getMass(string sym){
+
+    double mass = atomMap[sym].mass;
```
jkrajniak: This is not a clean approach. Why not to use ```checkSymbol``` to see if ```sym``` is a valid symbol first. With this you are creating the entry in the ```atomMap``` with key ```sym``` to only say that it is invalid because the mass is ```0```.

src/libcsg/atomtable.cc
```diff
@@ -0,0 +1,91 @@
...
+    if(mass<=0.0){
+        string err_msg = "Invalid atom symbol in AtomTable::getMass. There is"
+                         " no atom with symbol "+sym;
+        throw invalid_argument(err_msg);
+    }
+    return mass;
+}
+
+int AtomTable::getAtomicNumber(string sym){
+    int num = atomMap[sym].atomicNumber;
```
jkrajniak: The same comment as for ```getMass```

src/libcsg/modules/io/pdbreader.h
jkrajniak: What is this ```convention``` ? I see it only in dlpolyreader/writer. And where those setters/getters are used?

src/libcsg/modules/io/pdbreader.cc
```diff
@@ -57,128 +61,383 @@ void PDBReader::Close()
...
+        throw std::runtime_error("number of beads in topology and trajectory differ");
+
+
+    ////////////////////////////////////////////////////////////////////////////////
+    // Sort data and determine atom structure, connect with top (molecules, bonds)
+    ////////////////////////////////////////////////////////////////////////////////
+
+    // Extra processing is done if the file is a topology file, in which case the atoms
+    // must be sorted into molecules and the bond interactions recorded
+    if(_topology)
```
jkrajniak, line 277: I suppose that this part of the code is working correctly. But I think you could avoid all those comments by using other data structures. In fact what you need is a list of atoms in every molecule (e.g. again some ```unordered_map<int, list<int> >```) and for every atom id the corresponding molecule id. (e.g. ```map<int, int>```).

src/libcsg/modules/io/pdbreader.h
```diff
@@ -23,39 +23,45 @@
...
+        void setFname(string name) { _fname = name; return; }
+        /// get the topology file name: <name>.pdb (convention: ".pdb"="FIELD")
+        string getFname()          { return _fname; }
 
         void Close();
-
     private:
+        string _fname;
```
jkrajniak: where do you use ```_fname``` ?

jkrajniak (changes requested):

src/libcsg/modules/io/pdbreader.h
JoshuaSBrown: Other than in the setter and getter in the header file it is not used. Should I remove it then? I'm not sure if this is used elsewhere or not.

src/libcsg/modules/io/pdbreader.cc
JoshuaSBrown: Hmm ok, you are exposing me to objects I am not as familiar with... this is very good. I will see what I can do.

src/libcsg/atomtable.cc
JoshuaSBrown: nope

src/libcsg/atomtable.cc
JoshuaSBrown: I come from a high performance computing background, so if my approaches are not always clean it is sometimes related. I need the atom mass, in the case that the symbol is fine if I take your approach I will have iterated through two vectors the first time to check if the symbol is correct the second time to get the mass. It is extra work... arguably negligible in this case I will take your approach.

src/libcsg/modules/io/pdbreader.cc
```diff
@@ -57,128 +61,382 @@ void PDBReader::Close()
...
+                for(auto atm_temp =molecule_atms[obsolete_mol].begin();
+                         atm_temp!=molecule_atms[obsolete_mol].end();
+                         atm_temp++){
+
+                    atm_molecule[*atm_temp]=chosen_mol;
+                }
+
+                // Splicing will remove atoms from the now obsolete molecule and place them
+                // on the chosen molecule. 
+                molecule_atms[chosen_mol].splice(molecule_atms[chosen_mol].end(),
```
jkrajniak, line 356: Very nice!

jkrajniak (approved):
LGTM

junghans (approved):

PR #238, opened 2017-12-21
drop support for gmx-5.1

### LINKED ISSUE #237: Drop support for gromacs-5.1
Gromacs-5.1 reached end-of-life: https://mailman-1.sys.kth.se/pipermail/gromacs.org_gmx-developers/2017-December/009940.html

so let's remove support for it, too!

junghans:
Fix #237


CMakeModules/FindGROMACS.cmake
```diff
@@ -22,51 +22,52 @@
...
+  if(FOUND_GMX_IS_DOUBLE_PRECISION)
     set(GROMACS_DEFINITIONS "-DGMX_DOUBLE=1")
-  elseif(FOUND_GMX_IS_SINGLE_PRECISION AND GROMACS_VERSION GREATER 51)
+  elseif(FOUND_GMX_IS_SINGLE_PRECISION)
     set(GROMACS_DEFINITIONS "-DGMX_DOUBLE=0")
-  elseif(FOUND_GMX_IS_DOUBLE_PRECISION)
-    set(GROMACS_DEFINITIONS "-DGMX_DOUBLE")
   elseif(NOT FOUND_GMX_IS_SINGLE_PRECISION)
-    message(FATAL_ERROR "Could not find neither gmx_is_single_precision nor gmx_is_double_precision in the gromacs library, that is very very strange, take look at the error message in ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log to find out what was going wrong. This most likely means that your gromacs version is too old, we need at least gromacs 5.1 !") 
+    message(FATAL_ERROR "Could not find neither gmx_is_single_precision nor gmx_is_double_precision in the gromacs library, that is very very strange, take look at the error message in ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log to find out what was going wrong. This most likely means that your gromacs version is too old, we need at least gromacs 2016 !")
```
JoshuaSBrown: Could not find. Neither gmx_is_single_precision nor gmx_is_double_precision is defined in the gromacs library, that is very very strange, take a look at the error message in ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log to find out what was going wrong. This most likely means that your gromacs version is too old, we need at least gromacs 2016!

JoshuaSBrown: This is the only thing that I would change the English could use a little tweak in the error message.

JoshuaSBrown (commented):
Well it looks fine to me, other than a small tweak to the error message. Though I will admit I'm not very familiar with cmake yet.

JoshuaSBrown:
Is there a particular reason we are no longer supporting older versions of GROMACS?

CMakeModules/FindGROMACS.cmake
junghans: What text would you suggest?

junghans:
The problem is that, once they become unmaintained by upstream, bugs won't get fixed and user run the risk of producing physically-wrong results. (User can still use an older version of VOTCA if they need to run an old version of gromacs. Just to put this in perspective Gromacs-5.1 was released August 14, 2015.

CMakeModules/FindGROMACS.cmake
JoshuaSBrown: What I have copied in the first comment.

JoshuaSBrown (approved):

PR #249, opened 2018-06-16
Adding test to compare output of csg_reupdate

junghans:
Part of votca/votca#71


JensWehner (approved):
works for me, but why is the example so large?

junghans:
Good point, let me shrink that a bit!

junghans:
@JensWehner: Done

src/tools/CMakeLists.txt
```diff
@@ -21,3 +21,13 @@ foreach(PROG csg_reupdate csg_map csg_dump csg_property csg_resample csg_stat cs
     set_tests_properties(${PROG}Help PROPERTIES LABELS "csg;tools;votca")
   endif(ENABLE_TESTING)
 endforeach(PROG)
+
+if(ENABLE_TESTING)
```
junghans, line 25: @JoshuaSBrown that is the part you want to look at.

PR #250, opened 2018-06-17
lammpswriter: initial commit

junghans:
Dump is nice because it is text-based and it supports force, which should be useful to add some test down the road.


JoshuaSBrown (changes requested):
Looks great to me, but you don't have any tests, could we build in a simple test?

junghans:
Good point, but I want to use `votca_compare` (votca/tools#43) for that.

JoshuaSBrown (approved):

JoshuaSBrown:
Sounds good, you're the best!

junghans:
@JoshuaSBrown, test added!

PR #253, opened 2018-07-05
Adding csg_stat tests

junghans:


JensWehner:
how was the reference data created?

junghans:
By me, running the test!

PR #254, opened 2018-07-18
Test lammpswriter

JoshuaSBrown:
This branch adds testing for the lammps trajectory writer and reader.


junghans (approved):

src/tests/test_lammpsreaderwriter.cc
```diff
@@ -0,0 +1,423 @@
...
+
+		outfile.close();
+
+		Topology top;
+
+   	// Make square box
+		matrix box;
+		// 10  0  0
+		//  0 10  0
+		//  0  0 10
```
junghans, line 149: Isn't it 3 3 10?

JoshuaSBrown:
@junghans Can you explain to me what the Box object is supposed to contain? I thought it was supposed to contain 3 vectors depicting the edges? Which is why I set a box of:
   10 0 0 
   0 10 0
   0 0 10

junghans:
My point was that below the comment, you set it to something else, namely:
1 1 10
2 2 10
3 3 10

JoshuaSBrown:
Oh I thought that I was simply specifying the indices...

JoshuaSBrown:
@junghans Do you have any insight into the problem travis keeps reporting? I can't figure out what the problem is based on the error message.

JoshuaSBrown:
Ignore the last comment, I think I spotted it.

PR #260, opened 2018-07-28
Added default value for has_force

### LINKED ISSUE #259: Topology does not set default value on _has_force
This is may result in an uninitialized call to the hasForce method. It was found when trying to write to lammps dump file format. 

@junghans  were you aware of this?

JoshuaSBrown:
Fix #259


JoshuaSBrown:
https://github.com/votca/csg/issues/259

junghans (approved):

Issue #261, opened 2018-07-30
Bead class has redundant methods


JoshuaSBrown:
Why does the bead class have vec& getVel() and vec& Vel() and vec& getPos() and vec& Pos(). Why on earth are there two methods the getVel checks a bool to see if the value is set or not but the Vel method doesn't... I don't understand what the reasoning is for this @ruehle?

ruehle:
getPos ist const, so you can use it even if bead is passed as const ref. Pos is not which allows you to change it. There is actually no reason to call them differently, but you should have 2 methods, one with const and one without.

Getvel is checked as there is not always a velocity. This depends on the trajectory. However we assume that there always must be a position.

Hope this answers your questions.

ruehle:
That the vel method doesn't check could be an error. Also as vel is read/write, it is not so obvious what to do. I don't really remember. Sorry

JoshuaSBrown:
Ok, thanks for the clarification. The lack of the asserts in the Vel(), Pos(), W() U() and F() methods seems to be an error then.

JoshuaSBrown:
Sorry for my exasperation.

JoshuaSBrown:
It seems though that passing out a reference might completely defeat the purpose of having a boolean indicating that the instance has a defined position, velocity etc... I would think it would be safer to pass out a copy instead. 

Well I guess as long as there is an assertion check to ensure the vec has been set before using the reference methods it doesn't matter.

JoshuaSBrown:
@junghans do you have any input? I would not normally care but seeing that I want to derive some classes off of this one I would like to be sure the bead class is air tight.

junghans:
I have no more information!

PR #262, opened 2018-07-31
Added assertion checks when using reference methods

JoshuaSBrown:
See discussion: https://github.com/votca/csg/issues/261


junghans:
`unit_test_lammpsreaderwriter` needs a fix!

PR #265, opened 2018-08-01
Bug interaction class does not define initial values

### LINKED ISSUE #264: Interaction class has ambiguous name method
Setting the name of the interaction is completely overwritten by setting the molecule or group or index of the interaction. Furthermore, default values are not associated with these integers meaning garbage is likely to be output each time without warning if all three are not explicitly set.

JoshuaSBrown:
Was finding when running valgrind that garbage values will be used without warning if one does not set some of the values. This was solved by placing asserts where needed to ensure the users do not do this. Also found that there was ambiguity in the class because the the internal name is created by using the group group index molecule id and index of the class. However there was also a method to directly change the name without updating these other values setting up the class instance to become out of date by using the setName method. Thus the setName method was removed as it was not used anywhere in the code base anyway according to a search in netbeans. The change in the way the name is printed should also not have a large impact as it is only printed in the output of two files in the comment classes. 

Fix #264

### DIFF
--- a/include/votca/csg/interaction.h
+++ b/include/votca/csg/interaction.h
@@ -83,7 +83,7 @@ public:
   virtual vec Grad(const Topology &top, int bead) = 0;
   int BeadCount() { return _beads.size(); }
   int getBeadId(int bead) {
-    assert(bead > -1 && bead < _beads.size());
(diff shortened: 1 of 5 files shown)


JoshuaSBrown:
@junghans do you want to look over this before I merge?

junghans (approved):

PR #266, opened 2018-08-02
Feature base bead

JoshuaSBrown:
This base class will be used in both the bead class and as the foundation for the molecule and atom classes.


JoshuaSBrown:
So this is a dumbed down version of the bead class. I will be using it to 1. refactor the bead class 2. refactor the atom class. 

If you have any suggestions for methods to remove that are not used in both or would make the base class less flexible now would be a good time to mention them. 

I do have a question for @JensWehner do you know what is used in the KMC module? If there is some sort of particle class it might also make since to model it off this base class as well. E.g. what class is used to model the polarons.

JoshuaSBrown:
This branch should only be merged after the pull request in tools is merged: https://github.com/votca/tools/pull/53

JensWehner:
KMC has a carrier class. 
https://github.com/votca/xtp/blob/master/include/votca/xtp/chargecarrier.h

Concerning all the classes, we still have a little bit the problem of which unit to use. The XTP QMpart uses bohr, whereas everything else uses nm.

JoshuaSBrown:
@JensWehner thanks, it doesn't appear that the carrier class has enough similarities to bother trying to refactor with the bead structure so I guess I'll won't touch that.

JensWehner:
okay.

JoshuaSBrown:
@jkrajniak I would like your opinion here. I have used the identity and Name classes which I defined in tools and inherited from them. I did this because I would like most of the classes to inherit the methods and properties of these objects and doing it this way will ensure the names of the methods are uniform across multiple classes. However, that said I do not want to end up with the diamond of death.

junghans (approved):

junghans (approved):

jkrajniak:
@JoshuaSBrown This approach looks very good and it is quite popular. For the diamond problem, you could try to use virtual inheritance approach.

JoshuaSBrown:
@jkrajniak thanks your vote is reassuring 👍.

JoshuaSBrown:
Addressing diamond dependency issue first here in identity and name classes: https://github.com/votca/tools/pull/58

PR #270, opened 2018-08-13
Feature base bead structure

JoshuaSBrown:
This bead structure class will allow easy structure recognition. It uses the graph algorithm in tools as the backbone. Thus one is able to compare molecules simply by looking at the structure. This will be useful both in the fragment and molecule classes but also any topology class.

Seeing as this will be a dependency to several classes it would be great if I could locate improvements to the design sooner rather than later. Any suggestions are of course welcome :).


src/libcsg/beadstructure.cc
```diff
@@ -0,0 +1,176 @@
...
+    graphUpToDate = true;
+  }
+}
+
+void BeadStructure::CalculateStructure_() {
+
+  InitializeGraph_();
+  if (!structureIdUpToDate) {
+    findStructureId<GraphDistVisitor>(*graph_);
+    structureIdUpToDate = false;
```
jkrajniak: shouldn't be `true` ?

JoshuaSBrown: Good catch!

PR #271, opened 2018-08-15
Refactor bead base bead

JoshuaSBrown:


JoshuaSBrown:
Need to update votca submodules.

PR #275, opened 2018-08-23
Update Boost Version

JensWehner:


junghans:
Why?

JensWehner:
we have updated boost in xtp and tools, why use a different version test for unittests

PR #279, opened 2018-09-14
Molecule refactor testing

JoshuaSBrown:
Refactored the molecule class to make use of the beadstructure as a base class, then I proceeded to remove duplicate functionality. 

The bead structure makes it possible to make comparisons between the structures of molecules. This way you can actually use the isStructureEquivalent function to compare molecules based on their structure. I have also attempted to add some testing.


src/tests/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ foreach(PROG
   test_interaction
   test_lammpsdatareader 
   test_lammpsdumpreaderwriter
+  test_molecule
```
junghans: There is already a test called `test_molecule` in xtp, so you will need to rename one of the two.

JoshuaSBrown: Good point any suggestions?

junghans: `csg_test_molecule` ?

JoshuaSBrown:
This breaks ctp.

JoshuaSBrown:
This branch was put on hold because of an error that is addressed here: https://github.com/votca/csg/pull/283

will need to propagate the fix up to this branch.

JoshuaSBrown:
This branch should not be merged with master at this time.

PR #283, opened 2018-10-08
Thread class update

JoshuaSBrown:
Memory problem was found to originate from from miss using the Thread class is the csgapplication class. This was addressed in tools by making the Run method private, and in csgapplication it was fixed by using the Start and Wait methods in place of the run method. A few other things were updated in the process as well, but they are aesthetic mostly.

### DIFF
--- a/src/libcsg/csgapplication.cc
+++ b/src/libcsg/csgapplication.cc
@@ -125,7 +125,7 @@ namespace votca {
             //out << "\n\n" << OptionsDesc() << endl;
         }
 
-        void CsgApplication::Worker::Run(void) {
+        void CsgApplication::Worker::_Run(void) {
             while (_app->ProcessData(this)) {
                 if (_app->SynchronizeThreads()) {
                     int id = getId();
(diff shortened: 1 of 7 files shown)


junghans (approved):

JensWehner (changes requested):
What does the pull request do exactly?

src/libcsg/csgapplication.cc
```diff
@@ -368,7 +368,8 @@ namespace votca {
 
 
                 } else {
-                    master->Run();
+                    master->Start();
+                    master->WaitDone();
```
JoshuaSBrown, line 372: @JensWehner This is where the problem is. Master is a Thread object from tools. The Run method is not meant to be called except internally by the the start method. Secondly because WaitDone was not called before the delete, the thread was being deleted before it had finished causing all sorts of errors.

JensWehner: ahh I see, really good catch. 

Thanks

JensWehner (approved):

PR #288, opened 2018-11-30
remove const parameter from return by value

JensWehner:
gives a compiler warning.

Why did we not catch that in the compilation?


junghans (approved):

junghans:
What gcc version is it?

junghans:
And it passed here: https://travis-ci.org/votca/csg/builds/460959180

JensWehner:
I got it in icc18.

JensWehner:
https://software.intel.com/en-us/articles/cdiag858

JensWehner:
https://software.intel.com/en-us/articles/cdiag858

schererc:
No, I'm too late. I never got a warning at this point. But, okay

PR #289, opened 2018-12-11
CHANGELOG.md: update

junghans:
@jens, you need to do the same for xtp.

Should we do that for tools as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,20 @@
 For more detailed information about the changes see the history of the [repository](https://github.com/votca/csg/commits/master).
 
 ## Version 1.5-dev
-* dropped support for gmx-5.0
+* added support for 3 body distribution
+* allow usage of _ in gmx options
+* support espresso-4.0
+* added LAMMPS data topology reader
+* added LAMMPS dump writer
+* added support for gmx-2019
+* move from gsl to Eigen3
+* added support for running gmx in multidir mode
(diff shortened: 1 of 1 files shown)


JoshuaSBrown (approved):

JoshuaSBrown:
I suppose we could do something similar for tools. I have added the beginnings of some graph functionality. We have also updated the histogram class + all the stuff @JensWehner did with switching to Eigen.

junghans:
Are you volunteering?

JensWehner:
I ll do the xtp part, if joshua is happy to do it for tools?

JoshuaSBrown:
I can give it a shot but I'm leaving for Christmas break tomorrow and won't have access to internet until Wednesday next week.

junghans:
@JoshuaSBrown, can you do a:
```
git remote update
git log --oneline origin/stable..origin/master
```
and check if I forgot anything.

JoshuaSBrown:
Does that display all the commit that are different between the two branches? The only thing that I can see that I contributed that is missing is the addition of the beadstructure class

 *  added bead structure class which allows the comparison of structures of bonded beads

junghans:
I just added things, which matter for the end user! Not sure if the bead structure change is actually visible.

JensWehner (approved):
LGTM, I cannot think of anything else.

JoshuaSBrown:
@junghans than I have nothing to add.

PR #290, opened 2018-12-22
Add tests and update csg boltzmann

JoshuaSBrown:
In an attempt to remove use of histogram and replace it with histogramnew without breaking the code it was necessary to setup a few unit tests for classes that made use of the histogram class. This merge is responsible for creating these tests which are located in the csg-boltzmann folder. 
In response to issue https://github.com/votca/tools/issues/47

### DIFF
--- a/src/csg_boltzmann/CMakeLists.txt
+++ b/src/csg_boltzmann/CMakeLists.txt
@@ -1,5 +1,4 @@
 file(GLOB CSG_BO_SOURCES *.cc)
-add_library(votca_csg_boltzmann ${CSG_BO_SOURCES})
 add_executable(csg_boltzmann ${CSG_BO_SOURCES})
 target_link_libraries(csg_boltzmann votca_csg)
 install(TARGETS csg_boltzmann RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
@@ -14,3 +13,6 @@ if (TXT2TAGS_FOUND AND BASH)
   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/csg_boltzmann.man DESTINATION ${CMAKE_INSTALL_MANDIR}/man1 RENAME csg_boltzmann.1)
   set_property(DIRECTORY APPEND PROPERTY ADDITIONAL_MAKE_CLEAN_FILES csg_boltzmann.help csg_boltzmann.t2t)
 endif(TXT2TAGS_FOUND AND BASH)
+
+list(REMOVE_ITEM CSG_BO_SOURCES "main.cc")
+add_library(votca_csg_boltzmann ${CSG_BO_SOURCES})
--- a/src/csg_boltzmann/CMakeLists.txt
+++ b/src/csg_boltzmann/CMakeLists.txt
@@ -1,4 +1,5 @@
 file(GLOB CSG_BO_SOURCES *.cc)
+add_library(votca_csg_boltzmann ${CSG_BO_SOURCES})
 add_executable(csg_boltzmann ${CSG_BO_SOURCES})
 target_link_libraries(csg_boltzmann votca_csg)
 install(TARGETS csg_boltzmann RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
--- a/src/csg_boltzmann/tabulatedpotential.cc
+++ b/src/csg_boltzmann/tabulatedpotential.cc
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright 2009-2011 The VOTCA Development Team (http://www.votca.org)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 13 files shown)


junghans:
Does that fix #155?

junghans:
And could we add a regression test for csg_boltzmann? I.e echoing some commands into the command and compare the output.

JoshuaSBrown:
The boltzmann constant used in the boltzmann inversion function in the tabulated potential has a hard coded value in what I believe is units of [ kJ/(mol K) ]. Is this what you are referring too?

JoshuaSBrown:
It is not really that clear to me how to use csg_boltzmann. @junghans in the interactive mode I can specify a histogram etc... but is there a way to do this without going into the interactive mode? I'm also not sure what to use for the selection to get meaningful output? I've been trying with files in tools/references/spce but I don't know if this is appropriate either. 

    list: list all available bonds
    vals <file> <selection>: write values to file
    hist <file> <selection>: create histogram
    tab <file> <selection>: create tabulated potential
    autocor <file> <selection>: calculate autocorrelation, only one row allowed in selection!
    cor <file> <selection>: calculate correlations, first row is correlated with all other rows

junghans:
I was thinking about something like `echo "command1\ncommand2\nq" | csg_boltzmann`.

junghans:
I have did something like:
```
hist end.txt end-end
vals end-v.txt end-end
hist AA.txt AA-bond
hist AAA.txt AAA-angle
hist set periodic 1
hist set auto 0
hist set min -3.14159
hist set max 3.14159
hist AAAA.txt AAAA-dihedral
tab set T 400
tab U_AA.txt AA-bond
tab U_AAA.txt AAA-angle
tab set periodic 1
tab set smooth1 1
tab set auto 0
tab set min -3.14159
tab set max 3.14159
tab U_AAAA.txt AAAA-dihedral
autocor end-cor.txt end-end
autocor AA-cor.txt AA-bond[4]
autocor AAA-cor.txt AAA-angle[5]
autocor AAAA-cor.txt AAAA-dihedral[4]
cor bond-angle.txt AAA-angle[4] AA-bond
cor bond-dihe.txt AAAA-dihedral[4] AA-bond
cor angle-dihe.txt AAAA-dihedral[4] AAA-angle
q
```
and then `cat "cmd" | csg_boltzmann --top topol.tpr` before.

However it seems CMake has a hard time to redirect stdin into tests (see https://stackoverflow.com/questions/36304289/how-to-use-redirection-in-cmake-add-test).

So @JoshuaSBrown how hard would it be to add an option (e.g. `--in` to have `csg_boltzmann` read from a file instead of stdin?

JoshuaSBrown:
I can do that but let's put it in a separate merge, that will be more than a quick fix, this merge is good to go unless you see something else that I have done that should be changed?

junghans:
If you do that than I will write the tests!

src/csg_boltzmann/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 file(GLOB CSG_BO_SOURCES *.cc)
+add_library(votca_csg_boltzmann ${CSG_BO_SOURCES})
```
junghans: Won't that library have a `main()` if we use all `csg_boltzmann` sources?

junghans:
Will this help with https://github.com/votca/tools/issues/47?

junghans:
Maybe we can add an option to change the unit of `kB` down the road?

src/csg_boltzmann/CMakeLists.txt
JoshuaSBrown: Yes, I guess I could remove the main.cc that would be less sloppy.

JoshuaSBrown:
Yes, that could be a separate issue. How units are handled could be streamlined across the whole VOTCA package. Though I don't yet have a good feel for that aspect of the VOTCA package yet as I have not really used the tools with the various file formats we support.

src/csg_boltzmann/tabulatedpotential.cc
```diff
@@ -15,333 +15,348 @@
...
-#include "tabulatedpotential.h"
+#include <votca/tools/histogram.h>
 
 using namespace std;
 using namespace boost;
+using namespace votca::tools;
+using namespace votca::csg;
+
+/* Boltzmann constant in units [ kJ/(mol K) ] */
+const double kB = 0.0083109;
```
JensWehner: move to votca/tools/constants

src/csg_boltzmann/tabulatedpotential.cc
```diff
@@ -15,333 +15,348 @@
...
+
+  _min = numeric_limits<double>::max();
+  _max = numeric_limits<double>::min();
+
+  for (size_t i = 0; i < data.size(); i++) {
+    _max = max(data[i], _max);
+    if (data[i] > 0)
+      _min = min(data[i], _min);
+  }
+  _max = -kB * _Temperature * log(_max);
```
JensWehner: really good catch, well done

JensWehner (approved):
lgtm nice rework.  So we can remove histogram and rename histogram_new to histogram?

src/csg_boltzmann/tabulatedpotential.cc
JoshuaSBrown: Well you know I try... :)

src/csg_boltzmann/tabulatedpotential.cc
JoshuaSBrown: Sounds good.

JoshuaSBrown:
@JensWehner and yes that is the idea, once we have good tests in place we should be able to replace histogram with histogramnew and then remove it hopefully without any hiccups.

JoshuaSBrown:
https://github.com/votca/tools/pull/84 should be merged first.

PR #296, opened 2018-12-30
cmake: add REGRESSIONTEST_TOLERANCE option

### LINKED ISSUE #295: make it possible to change the default precision for regression tests via CMake
On some archs the default tolerance of 1e-5 is a bit too small, e.g. on aarch64 we need 2e-5.

junghans:
Fix #295


JensWehner (approved):
LGTM

PR #297, opened 2019-01-03
Several source files that depended on using namespace std from tools were corrected. 

JoshuaSBrown:
Full scope std:: was added to these files.


JoshuaSBrown:
Do not merge yet, I have more commits I need to push.

PR #300, opened 2019-01-07
Stable

JoshuaSBrown:
I don't know if this is the correct way to do this, but I'm guessing we want to pull in the bug fixes from stable into master?


junghans:
There are a bunch of conflicts here!

JoshuaSBrown:
So what is the normal procedure, do you cherrypick?

junghans:
No, merging is right, you just need to resolve the conflicts.

junghans:
Is it never good to use the stable branch directly, see #301.

Leave this PR open for now.

Issue #302, opened 2019-01-08
csg_stat: Topology does not have beads of type "ME"


junghans:
After merging #298, I get the following error in when running the `methanol-water_X_0.062_simplex` repression test:
```
csg_stat --nt 4 --options /home/christoph/votca/src/votca/build/csg-tutorials/methanol-water/X_0.062/re/settings.xml --top topol.tpr --trj traj.xtc --begin 0 --first-frame 0
begin to calculate distribution functions
# of bonded interactions: 0
# of non-bonded interactions: 3
Reading file topol.tpr, VERSION 2019-rc1-dev-20181217-b3b270485-dirty-unknown (double precision)
I have 4000 beads in 4000 molecules
Reading file topol.tpr, VERSION 2019-rc1-dev-20181217-b3b270485-dirty-unknown (double precision)
Reading file topol.tpr, VERSION 2019-rc1-dev-20181217-b3b270485-dirty-unknown (double precision)
Reading file topol.tpr, VERSION 2019-rc1-dev-20181217-b3b270485-dirty-unknown (double precision)
Reading frame       0 time    0.000
an error occurred:
Topology does not have beads of type "ME"
This was specified in type1 of interaction "ME-ME"
```

JoshuaSBrown:
Um, how do you create the .xtc file?

junghans:
Just use `conf.gro` instead of `traj.xtc`. `topol.tpr` can be created by running `gmx grompp`.

junghans:
Fixed in #305

PR #298, opened 2019-01-04
Bug beadtype memory leak

JoshuaSBrown:
If I fix the memory leak here, the integration tests fail. 

      Start 14: Run_csg_fmatch
14/50 Test #14: Run_csg_fmatch .........................***Exception: SegFault  0.23 sec
      Start 15: Compare_csg_fmatch_output
15/50 Test #15: Compare_csg_fmatch_output ..............***Failed    0.03 sec
      Start 16: Run_csg_fmatch_3body
16/50 Test #16: Run_csg_fmatch_3body ...................***Exception: SegFault  0.23 sec
      Start 17: Compare_csg_fmatch_3body_output1
17/50 Test #17: Compare_csg_fmatch_3body_output1 .......***Failed    0.03 sec
      Start 18: Compare_csg_fmatch_3body_output2
18/50 Test #18: Compare_csg_fmatch_3body_output2 .......***Failed    0.03 sec
      Start 19: Compare_csg_fmatch_3body_output3
19/50 Test #19: Compare_csg_fmatch_3body_output3 .......***Failed    0.03 sec
      Start 20: Run_csg_map
20/50 Test #20: Run_csg_map ............................   Passed    0.03 sec
      Start 21: Compare_csg_map_output
21/50 Test #21: Compare_csg_map_output .................***Failed    0.05 sec
      Start 22: Run_csg_stat
22/50 Test #22: Run_csg_stat ...........................***Exception: SegFault  0.22 sec
      Start 23: Compare_csg_stat_output1
23/50 Test #23: Compare_csg_stat_output1 ...............***Failed    0.03 sec
      Start 24: Compare_csg_stat_output2
24/50 Test #24: Compare_csg_stat_output2 ...............***Failed    0.03 sec
      Start 25: Run_csg_stat_angular
25/50 Test #25: Run_csg_stat_angular ...................***Exception: SegFault  0.23 sec
      Start 26: Compare_csg_stat_angular_output
26/50 Test #26: Compare_csg_stat_angular_output ........***Failed    0.03 sec
      Start 27: Run_csg_stat-imc
27/50 Test #27: Run_csg_stat-imc .......................***Exception: SegFault  0.23 sec
      Start 28: Compare_csg_stat-imc_output
28/50 Test #28: Compare_csg_stat-imc_output ............***Failed    0.03 sec
      Start 29: Compare_csg_stat-imc_output_2
29/50 Test #29: Compare_csg_stat-imc_output_2 ..........***Failed    0.03 sec

PR #303, opened 2019-01-09
added csg_dump test

junghans:
Related to #302 , test passed locally before #298.

### DIFF
--- a/include/votca/csg/basebead.h
+++ b/include/votca/csg/basebead.h
@@ -70,6 +70,13 @@ public:
   virtual std::weak_ptr<BeadType> Type() const { return _type; }
 
   /**
+   * get the name of the bead type
+   * \return - string indicates the name or throw error if bead type is not 
+   * accesible.
+   **/
+  std::string getBeadTypeName();
+
+  /**
    * get the mass of the base bead
    * \return - base bead mass
    */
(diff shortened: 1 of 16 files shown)


junghans:
It passed locally! Now let's merge it with stable (which includes #298) and we get a segfault!

JoshuaSBrown:
This branch should be built with: https://github.com/votca/csgapps/pull/14

junghans:
@JensWehner, is there a way to avoid the duplicated:
```
throw runtime_error("Error in dlpolytopologyreader.cc in trying to access bead type it does not exist.");
```

JensWehner (commented):
I am not sure. This solution is rather hacky, as we do not yet understand the underlying cause. I do not know, I would return to it as a last resort. I also do not like all the error message, but I do not know what code changes would be necessary to fix this properly. I do not know. I ll look in the morning again

JensWehner:
@junghans we can put the error into a function of the baseclass, to have it only once.

junghans:
That sounds better!

JoshuaSBrown:
I think part of the problem is the method name. GetOrCreateBeadType() I've created an issue here. I think refactoring the methods would help clarify how this method is being used:
https://github.com/votca/csg/issues/305

JensWehner:
Concerning the use of std::runtime_error, errors should not be used to check for incorrect usage of code in my mind, but for wrong data. Asserts are for that.

JoshuaSBrown:
@junghans would it be a good idea to setup travis to automatically build with branches that have the same name?

junghans:
We had builds for all branches enabled in the past, but it triggered far too many builds! And without branch builds it is quite hard to figure out if there are two pull requests open pointing to branches with the same name!

Also this case happens only in very few cases and you can always triggered it manually!

PR #304, opened 2019-01-10
Stable fix beadtype segfault

JoshuaSBrown:
@junghans your test is implemented here already. By switching to shared and weak pointers for the beadtype, the tests pass locally on ubuntu. I'm still unsure as to what the underlying cause is though. Let me check if valgrind is still finding problems before we merge. I would also like to cleanup the code a bit before.


junghans:
Leave it open for now.

PR #310, opened 2019-01-25
Cleaned up multiple inheritance in base bead

JoshuaSBrown:
1. Changed from multiple inheritance to composition using name and identity objects
2. Replaced bead and basebead variable names to more explicit names
3. Replaced prepended underscore to post pended underscore to satisfy the standard.


include/votca/csg/basebead.h
```diff
@@ -110,40 +128,46 @@ class BaseBead : public TopologyItem,
...
+  bead_position_set_ = true;
+  bead_position_ = bead_position;
 }
 
 inline const vec &BaseBead::getPos() const {
-  assert(_bPos);
-  return _pos;
-}
-}
+  assert(bead_position_set_);
```
JensWehner: make it `assert(bead_position_set_ & "Position not set")` an assert without a message is really hard to debug

include/votca/csg/basebead.h
```diff
@@ -110,40 +128,46 @@ class BaseBead : public TopologyItem,
    * \return reference to position
    */
   virtual vec &Pos() {
-    assert(_bPos);
-    return _pos;
+    assert(bead_position_set_);
```
JensWehner: see below

include/votca/csg/bead.h
```diff
@@ -186,51 +186,51 @@ class Bead : public BaseBead {
    * \return reference to velocity
    */
   vec &Vel() {
-    assert(_bVel);
-    return _vel;
+    assert(bead_velocity_set_);
```
JensWehner: see above

include/votca/csg/bead.h
```diff
@@ -186,51 +186,51 @@ class Bead : public BaseBead {
...
   }
 
   /**
    * direct access (read/write) to orientation u of the bead
    * \return reference to u
    */
   vec &U() {
-    assert(_bU);
-    return _u;
+    assert(bU_);
```
JensWehner: again

include/votca/csg/bead.h
```diff
@@ -276,7 +276,7 @@ class Bead : public BaseBead {
    * If it is a mapped beads, returns te bead id the cg bead was created from
    * \return vector of bead ids of reference atoms
    */
-  std::vector<int> &ParentBeads() { return _parent_beads; };
+  std::vector<int> &ParentBeads() { return parent_beads_; };
```
JensWehner: why not const and if not create also a const version

include/votca/csg/bead.h
```diff
@@ -308,49 +312,49 @@ class Bead : public BaseBead {
...
-    _bF = false;
+    bead_position_set_ = false;
+    bead_velocity_set_ = false;
+    bU_ = false;
+    bV_ = false;
+    bW_ = false;
+    bead_force_set_ = false;
   }
 
   void *_userdata;
```
JensWehner: where is this used? The type information has been cased away

JensWehner: @junghans do you have an idea about this. I am not good with pointers

JensWehner (changes requested):
good step in the right direction

include/votca/csg/bead.h
JoshuaSBrown: I keep seeing this user data pointer in other classes but it still isn't clear to me what exactly they are used for.

JoshuaSBrown: I can't find it anywhere so I guess I'll pull it out.

JoshuaSBrown: The same is true for Options it does not appear to be used anywhere...

JoshuaSBrown:
I removed the options and user data from the beads, nothing is ever added to these attributes they are just passed around. I imagine the intention was that if someone ever wanted to extend the functionality without making changes to the underlying structures they could used these attributes.

JoshuaSBrown:
@JensWehner do you want to check this once more before I merge?

PR #311, opened 2019-01-25
Remove bead type

JoshuaSBrown:
This code removes the malloced BeadType and replaces it with a string.


JoshuaSBrown:
This should be merged first:
https://github.com/votca/csg/pull/310

JoshuaSBrown:
Most of the code that was changed is a result of the clang formatter. The functions that were changed are those associated with the topology object. 


Methods added are:
```
  /**                                                                            
   * \brief Determine if a bead type exists.                                     
   *                                                                             
   * @return bool true if it has been registered                                 
   **/                                                                           
  bool BeadTypeExist(std::string type) const;                                    
                                                                                 
  /**                                                                            
   * \brief Register the bead type with the topology object.                     
   *                                                                             
   * Records are kept of the different bead types in the topology object. This   
   * method stores the bead type.                                                
   **/                                                                           
  void RegisterBeadType(std::string name);      
```

PR #323, opened 2019-01-31
introduced more templates to make filereaders more general

JensWehner:
@junghans @JoshuaSBrown 

This is my idea of the more general filereaders/writers. Now any class can write and read xyz files as long as it has certain functionality. 

For an example see: https://github.com/votca/xtp/blob/electrostatics/src/libxtp/qmmolecule.cc

Disadvantages:
- Atm it expects all containers but topology to use bohr as the unit, but I have an idea how to fix it if it comes up at some point
- XYZ reader and writer header files are more visible now, as I cannot go through the factory, we can also template the factory but atm I do not think that is necessary
- The template specialisation is a bit messy, but if we at some point clean up the csg topology a bit more that should go away

- Same was done for the .pdb writer. Atm I do not want to expand it further only if necessary at some point. 

I ll try the pdb reader next and see what I can do. .xyz and .pdb are the most common readers in XTP so replacing them would go a long way


include/votca/csg/xyzreader.h
```diff
@@ -0,0 +1,150 @@
...
+        Tokenizer tok1(line," \t");
+        std::vector<std::string> line1;
+        tok1.ToVector(line1);
+        if(line1.size()!=1){
+          throw std::runtime_error("First line of xyz file should contain number of atoms/beads, nothing else.");
+        }
+        int natoms = boost::lexical_cast<int>(line1[0]);
+        if(!topology && natoms !=getContainerSize(container))
+            throw std::runtime_error("number of beads in topology and trajectory differ");
+
```
JoshuaSBrown: Same as above.

include/votca/csg/xyzreader.h
```diff
@@ -0,0 +1,150 @@
...
+        // the title line
+        getline(_fl, line); ++_line;
+
+        // read atoms
+        for(int i=0; i<natoms; ++i) {
+            getline(_fl, line); ++_line;
+            if(_fl.eof())
+                throw std::runtime_error("unexpected end of file in xyz file");
+            vector<string> fields;
```
JoshuaSBrown: Same

include/votca/csg/xyzreader.h
```diff
@@ -0,0 +1,150 @@
...
+            Tokenizer tok(line, " ");
+            tok.ToVector(fields);
+
+            if(fields.size() != 4)
+                throw std::runtime_error("invalide line " +
+                        boost::lexical_cast<string>(_line) +
+                        " in xyz file\n" + line);
+
```
JoshuaSBrown: Same

src/libcsg/modules/io/xyzreader.cc
```diff
@@ -18,8 +18,7 @@
 #include <vector>
 #include <boost/lexical_cast.hpp>
 #include <votca/tools/getline.h>
-#include "xyzreader.h"
-
+#include <votca/csg/xyzreader.h>
```
JoshuaSBrown, line 21: I think that local includes are always preferred actually.

include/votca/csg/xyzreader.h
```diff
@@ -0,0 +1,150 @@
...
+        void AddAtom(Topology &container,std::string name,int id, const tools::vec& pos){
+        Bead *b;
+        tools::vec posnm=pos*tools::conv::ang2nm;
+        if(topology)
+            b = container.CreateBead(1, name+boost::lexical_cast<string>(id),
+                    (container.GetOrCreateBeadType(name)), 0, 0, 0);
+        else
+            b = container.getBead(id);
+            b->setPos(posnm);
+        }
```
JoshuaSBrown: Please put brackets around if { } else and don't rely on indents. I have introduced several bugs by not paying attention to indents. I guess you just copied the previous code might as well make it safer.

JoshuaSBrown (changes requested):
It looks very clean, much better than the current implementation. I say we use your approach.

include/votca/csg/xyzreader.h
JensWehner: ??

include/votca/csg/xyzreader.h
JensWehner: ?

include/votca/csg/xyzreader.h
JensWehner: ?

src/libcsg/modules/io/xyzreader.cc
JensWehner: okay, but it is in a different directory now, so I had to change the includes

include/votca/csg/xyzreader.h
JensWehner: okay

src/libcsg/modules/io/xyzreader.cc
JoshuaSBrown: ok

include/votca/csg/xyzreader.h
JoshuaSBrown: My comments appeared in a different order. The point is to not rely on indentation but explicitly put brackets around the content in if statements or loops etc...

JensWehner: ahh.

JoshuaSBrown (approved):

junghans (changes requested):
It seems you didn't use `git mv` to move the header to a new location hence the history is broken! Can you do a rebase to fix that?

Plus, I would put the header in `include/votca/csg/reader` instead!

junghans (approved):
Or maybe the files are just too different!

junghans:
I still don't like to have the implementation in the header itself, I/O never so performance critical that you couldn't call a library function.

JensWehner:
@junghans unfortunately it has to be in the header, because it is a template. The template has to been seen by XTP, so that XTP can create the code for the XTP class from the CSG template. So it must be in the header. If that is a big minus for you, we have to think of a different solution

junghans:
Well, makes sense then, I guess!

JensWehner:
@junghans @JoshuaSBrown  This can be merged from my side

junghans:
Something is wrong here: ede15c46e6e94cf1dab6d64ca4de91f9d38a877a is called **merged**, but it is not a merge!

junghans (approved):
It seems @JensWehner and @JoshuaSBrown  are using different version of `clang-format`, we should agree on one (e.g. the one in Fedora) to avoid flip-flop between the two versions.

JensWehner:
yeah, any idea how?

JensWehner:
should we maybe put a clang-format executable in the repo?

junghans:
We could check the version in the hook and only run formatting when the right version is presented!

JensWehner:
yeah that would also work.

JensWehner:
more discussion of clang-format


https://github.com/votca/dev-tools/issues/5

include/votca/csg/basebead.h
```diff
@@ -148,7 +146,7 @@ inline const TOOLS::vec &BaseBead::getPos() const {
          "Cannot get bead position as it has not been set.");
   return bead_position_;
 }
-}  // namespace csg
-}  // namespace votca
+} // namespace csg
```
junghans, line 149: @JensWehner @JoshuaSBrown just FYI, these whitespace changes in comments are due to a different clang-format version!

JensWehner: okay so what do we do now?

junghans: Fixing yourselves to one version of clang-format!

JoshuaSBrown: @JensWehner I'm using: clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
what version is yours?

JensWehner:
clang-format version 3.8.0-2ubuntu4

PR #324, opened 2019-02-01
[WIP] Joshs development

JoshuaSBrown:
Should only be merged after:
[remove bead type branch is merged](https://github.com/votca/csg/pull/311)
[base bead remove multiple inheritance branch is merged](https://github.com/votca/csg/pull/310)
[Jens file reader branch is merged](https://github.com/votca/csg/tree/filereaders)

1. This branch adds a new class base molecule, base molecule shared methods that are common both to the csg molecule and the xtp molecule classes, it also inherits from the beadstructure class which provided structural identification functionality. 

2. The local csg molecule was made to inherit from the base molecule. 

3. The base molecule and the beadstructure were converted to templated classes to provide a higher level of flexibility. 

4. The residue class was removed, as it was not needed and was a known source of memory leaks. 

5. Following functions were removed from topology.h file as they were not used anywhere, and their implementations were faulty:
 * void CreateMoleculesByResidue();
 * void CreateOneBigMolecule(std::string name);
 * void Add(Topology *top);

6. Separate from the residue name beads now also require the molecule name to be passed in when calling their constructor. This was to prevent confusing the residue and molecule names when they are two distinctly different objects. 

7. The add bead methods were changed from:
```void AddBead(Bead *bead, const std::string &name);```
to
```void AddBead(T *bead);```

Because it was redundant to pass the name in as a separate variable when the name is already stored in the bead. 

8. Randomly throughout the file readers in csg the bead names are sometimes created with the following syntax:
"ResidueName:ResidueNumber:AtomName"

with colons separating the different fields. Instead the bead was provided with a new method called get label that will automatically generate the appropriate string based on the residue name number and atom name stored in the bead. 
9. Memory tests with valgrind were added to cmake

10. Memory errors were fixed.


junghans:
@JoshuaSBrown can we delete this branch?

PR #329, opened 2019-02-05
Format all files

JoshuaSBrown:
Applying clang format to all source and header files.

### DIFF
--- a/include/votca/csg/beadlist.h
+++ b/include/votca/csg/beadlist.h
@@ -1,5 +1,5 @@
-/* 
- * Copyright 2009-2018 The VOTCA Development Team (http://www.votca.org)
+/*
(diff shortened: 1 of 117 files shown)


JensWehner:
I thought we only run it in our hooks and then bit by bit do this?

JoshuaSBrown:
Ya, but it's creating problems when merging, its hard to tell what's been changed, when half the files have been formatted. This is better. The only thing that I had to change was I had to remove 

#include <votca/csg/potentialfunctions/potentialfunctioncbspl.h>

from 

src/libcsg/potentialfunctions/potentialfunctioncbspl.cc

JoshuaSBrown:
Now if we format everything on our branches before we merge it will be obvious what things have been updated.

PR #332, opened 2019-02-07
Stable pcc64le fix lexical cast

JoshuaSBrown:


junghans (changes requested):
For stable fixes:

- The branch needs to be names `/for/stable/...`
- The patches should be concise as they might need to be patched into tarball source, so  please no clang-format etc. as it will lead to a higher failure rate.

In general, with all the formatting changes this unreviewable, I have no idea what changed!

Just avoid cosmetic change for stable, this is too late for the code base anyhow! (For master that is fine.)

junghans:
@JoshuaSBrown can this be replaced by #333 ?

junghans:
Fixed by #333

PR #331, opened 2019-02-07
Pcc64le fix lexical cast

JoshuaSBrown:

PR #333, opened 2019-02-07
pcc64le: fix lexical cast

junghans:
Fix #331 
Fix #332


JoshuaSBrown:
So what is different about this merge than the ones I did before? Do you have to prepend with for/stable/ ? I see you already answered.

JoshuaSBrown:
Looks good to me!

junghans:
@JoshuaSBrown yes, to test against `stable` you need to name the branch `for/stable/....`!

PR #337, opened 2019-02-13
[WIP] Topology type container

JoshuaSBrown:
This branch comprises of several different changes: 
1. Topology class has been turned into a template
2. A new class has been introduced for managing the different types existing in the topological object:
 a. Bead types
 b. Molecule types
 c. Residue types
3. Interaction class has been altered to use the BoundaryCondition class in place of the topology class, was not otherwise possible to compile. . Standardization of the way molecules and beads are constructed has also been attempted. I have tried to preserve as much as possible the same atom/bead ids as is indicated in which ever file is being read as opposed to arbitrarily creating the ids on the fly with a for loop. The thought being that if you are doing it on the fly you risk assigning information to wrong atoms/beads  and subsequent reads may not correctly conserve atom and bead information. 

Beads now are constructed with additional information if the information is not available it is clearly indicated with variables specifically used to indicate that the property is unassigned. Atoms and beads also have an element attribute to indicate if they represent a single element or not. 

Classification of molecules, residues, bead/atoms, elements
**molecules** - consist of a combination of 1 or more residues, 1 or more beads/atoms that are covalently bonded
**residues** - residues are not used in any specific manner, the information is just stored to be consistent with what is being read in by the file readers, according to the literature a residue is not necessarily the same thing as a molecule, it is a subunit of a molecule which can consist of one or more atoms/beads.
**bead/atoms** - the basic unit can consist of one or more elements that have been coarse grained, may or may not have the same name as an element. 
**elements** - an attribute of an atom/bead indicating if it is an element and if so which element, this is not the same as the atom/bead type as atoms and beads could have different names for the same element e.g. C1 C2 

5. Removed back pointers of molecule and topology objects, 
6. Finally, bead/atom name and type have been consolidated. Type has typically been an integer value but really conveys the same information as the name, as such the bead/atom name method has been removed and instead a bead/atom type method that accepts a string has been used instead.


include/votca/csg/bead.h
```diff
@@ -368,8 +371,9 @@ class Bead : public BaseBead {
 
   // void *_userdata;
 
-  friend class Topology;
-  friend class Molecule;
+  //  friend class Molecule;
+  friend class CSG_Topology;
```
JensWehner, line 363: leave name as topology, it is in a different namespace and thus enough

include/votca/csg/exclusionlist.h
```diff
@@ -51,7 +57,24 @@ class ExclusionList {
     std::list<Bead *> _exclude;
```
JensWehner, line 51: can we make this a vector, because I think, these are non owning pointers anyway

include/votca/csg/interaction.h
```diff
@@ -35,12 +36,13 @@ namespace csg {
 
     \todo double names/groups right, add molecules!!
 */
+
 class Interaction {
  public:
   Interaction() : _index(-1), _group(""), _group_id(-1), _name(""), _mol(-1){};
```
JensWehner: move this to the member variable list

include/votca/csg/map.h
```diff
@@ -33,15 +33,16 @@ class BeadMap;
...
 
   void AddBeadMap(BeadMap *bmap) { _maps.push_back(bmap); }
 
   void Apply();
 
  protected:
-  Molecule _in, _out;
+  const Molecule molecule_in_;
+  Molecule bead_out_;
   std::vector<BeadMap *> _maps;
```
JensWehner: does map own this, maybe then simply make it a vector<beatmap>

include/votca/csg/topologymap.h
```diff
@@ -29,21 +29,21 @@ class TopologyMap {
...
 
   void Apply();
 
  private:
-  Topology *_in;
-  Topology *_out;
+  CSG_Topology *_in;
+  CSG_Topology *_out;
 
   typedef std::vector<Map *> MapContainer;
```
JensWehner: maybe make this a std::vector<map>

src/libcsg/beadlist.cc
```diff
@@ -67,17 +76,21 @@ int BeadList::GenerateInSphericalSubvolume(Topology &top, const string &select,
...
-        push_back(*iter);
+      if (wildcmp(pSelect.c_str(), bead_ptr->getType().c_str())) {
+        push_back(bead_ptr);
       }
     } else {
-      if (wildcmp(pSelect.c_str(), (*iter)->getName().c_str())) {
-        push_back(*iter);
-      }
+      throw runtime_error("Bead name is no longer supported");
+      // if (wildcmp(pSelect.c_str(), bead_ptr->getName().c_str())) {
```
JensWehner, line 87: do you think this is a problem?

src/libcsg/cgengine.cc
```diff
@@ -39,17 +39,18 @@ CGEngine::~CGEngine() {
 /**
     \todo melts with different molecules
 */
-TopologyMap *CGEngine::CreateCGTopology(Topology &in, Topology &out) {
-  MoleculeContainer &mols = in.Molecules();
-  MoleculeContainer::iterator iter;
+TopologyMap *CGEngine::CreateCGTopology(CSG_Topology &in, CSG_Topology &out) {
```
JensWehner: not return a pointer

JensWehner: raw pointer is fine, but the problem is at the moment we are not yet consistent so we do not know if a raw pointer is owning or not

src/libcsg/map.cc
```diff
@@ -47,10 +47,11 @@ void Map::Apply() {
   }
 }
 
-void Map_Sphere::Initialize(Molecule *in, Bead *out, Property *opts_bead,
+void Map_Sphere::Initialize(const CSG_Topology *topology_parent_,
```
JensWehner: would the pbc object not be enough?

src/libcsg/modules/io/dlpolytopologyreader.cc
```diff
@@ -224,6 +227,17 @@ bool DLPOLYTopologyReader::ReadTopology(string file, Topology &top) {
         if (!top.BeadTypeExist(beadtype)) {
           top.RegisterBeadType(beadtype);
         }
+
+        Elements elements;
```
JensWehner: elements object should be created outside for loop

src/libcsg/modules/io/groreader.cc
```diff
@@ -116,8 +119,22 @@ bool GROReader::NextFrame(Topology &top) {
       }
 
       // res -1 as internal number starts with 0
-      b = top.CreateBead<Bead>(1, atName, atName, residue_number - 1, resName,
-                               resName, 1., 0.);
+      byte_t symmetry = 1;
```
JensWehner, line 119: why this is all needed?

src/libcsg/modules/io/growriter.cc
```diff
@@ -60,7 +60,7 @@ void GROWriter::Write(Topology *conf) {
     string resname =
         top->getBead(i)
             ->getResidueName();  // top->getResidue(resnr)->getName();
-    string atomname = top->getBead(i)->getName();
+    string atomname = top->getBead(i)->getType();
```
JensWehner, line 74: what does this change do?

JensWehner (commented):
where is the definition of topology?

JensWehner:
YOu forgot to add your headers.. I am also not sure if the template names are good, if you have to template topology call it topology_template or so

JoshuaSBrown:
@JensWehner I have a bunch of constants I have defined for attributes that are not defined, e.g. in the basebead_constants namespace the molecules_constants namespace and the bead_constants  namespace. This implementation I feel is not ideal do you have better suggestion?

JoshuaSBrown:
Take a look at **test_nblist_3body.cc** for an implementation.

JoshuaSBrown:
The **molecule_constants**, **bead_constants**, and **basebead_constants** are all in different header files. basebead.h, bead.h and molecule.h.

src/libcsg/modules/io/growriter.cc
JoshuaSBrown: I have gotten rid of all the setName and getName methods because they are ambiguous with the ids and the types of each object. They were simply being used like this 
Atom Type Carbon Id 1 

Name was being assigned something like this:
"Carbon#1" 
Which is redundant with the id and does not add anything useful.

JensWehner:
yeah but why?

JoshuaSBrown:
Why to which part? Why create constants in the first place or why are they all in different files?

JoshuaSBrown:
Constants were created in the first place to provide a means to explicitly describe attributes that were not defined. For instance molecule_id_unassigned, clearly indicates that the molecule id has not been assigned and avoids the use of some random magic number. This is more obvious for instance with the residue_type, previously the different file parsers were assigning different strings when the residue type was unknown e.g. "DUM" or "RES", there was no well defined way of dealing with them. 

They are all in different files, because I placed them with the objects that would use them, though in hindsite I think it would be better to define them in a single file. What are your thoughts?

JensWehner:
single file. The rest I understand.

src/libcsg/cgengine.cc
JoshuaSBrown: Raw pointer should never be owning... I think.

include/votca/csg/topologymap.h
JoshuaSBrown: ? You mean get rid of the typedef?

include/votca/csg/map.h
JoshuaSBrown: Yes, it does I have make it a vector of unique_pointers now, I cannot use the raw type because it makes use of polymorphism.

include/votca/csg/interaction.h
JoshuaSBrown: I don't follow?

include/votca/csg/exclusionlist.h
JoshuaSBrown: I don't care. I'll give it a shot.

include/votca/csg/bead.h
JoshuaSBrown: Take a look at templatetopology.h and csgtopology.h, I think this will work better.

src/libcsg/map.cc
JoshuaSBrown: Yes, I have now changed it.

src/libcsg/modules/io/dlpolytopologyreader.cc
JoshuaSBrown: Done.

src/libcsg/modules/io/groreader.cc
JoshuaSBrown: I am not a fan of hidden default variables. If a developer is forced to explicitly describe each attribute they will pay more attention to what they are doing. Though there is probably a better way to pass these values in as opposed to having a long list of parameters.

PR #346, opened 2019-02-26
more unit tests for interaction

JensWehner:
added unit tests for bond, angle, diheadral

### DIFF
--- a/src/tests/test_interaction.cc
+++ b/src/tests/test_interaction.cc
@@ -34,8 +34,8 @@ using namespace votca::csg;
 BOOST_AUTO_TEST_SUITE(interaction_test)
 
 BOOST_AUTO_TEST_CASE(test_interaction_constructor) {
-  IBond     bond1(1, 2);
-  IAngle    angle1(1, 2, 3);
+  IBond bond1(1, 2);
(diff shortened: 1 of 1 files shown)


JoshuaSBrown (commented):
What's up with the formatting, the clang-format file should not be aligning the = as per our discussion a few months ago.

JoshuaSBrown (commented):
Just a heads up I recently found out that calling endl calls flush every time and that it is preferred to call "\n" explicitly in place of std::endl; unless you are really wanting to flush the buffer.

JoshuaSBrown (changes requested):
Figure out what is wrong with the formatting.

junghans (approved):

junghans:
@JoshuaSBrown for a unit test the flushing of the buffer doesn't matter much, plus the output of tests gets redirected to a file with through buffering the filesystem anyhow.

JoshuaSBrown:
Yes, I it doesn't matter in this case, but I found that out the other day and thought it was a useful thing to know. You could potentially take a performance hit when outputting a lot of data... maybe.

junghans:
So can we merge this?

JoshuaSBrown (approved):

PR #357, opened 2019-03-31
FindGROMACS.cmake: add more useful error messages

junghans:


junghans:
Travis doesn't pick up base branch changes, so #359

PR #366, opened 2019-04-07
Added missing prepended integration_

JoshuaSBrown:
CMakeLists.txt DEPEND command was not updated with the correct filenames after prepending with integration_.


junghans (approved):

junghans:
Maybe we should also use `LABELS` more, as that allows `ctest -L <LABEL>`.

JoshuaSBrown:
That seems like a more robust option than globbing.

PR #368, opened 2019-04-09
Target link lib

JensWehner:
Reduces scope of many library symbols


junghans (dismissed):
Just out of curiosity, why can votca_tools not be private?

JensWehner:
I forgot, but either a unittest or one of the calculator needed symbols.

junghans:
Ok

junghans (approved):

PR #378, opened 2019-05-18
CMake Refactor in csg

junghans:
Part of https://github.com/votca/votca/pull/160


CMakeLists.txt
```diff
@@ -69,31 +62,18 @@ else (BUILD_MANPAGES)
 endif (BUILD_MANPAGES)
 
 find_package(Boost 1.53.0 REQUIRED COMPONENTS program_options filesystem system )
```
JoshuaSBrown, line 64: Do we need to add every boost header that we use here, because if that is the case this line is out of date?

JensWehner: I think so

junghans: @JoshuaSBrown, yes, we should add every package we are using here. Which ones are missing?

PR #392, opened 2019-08-13
Electrostatics

JoshuaSBrown:
Update my development branch with Jens recent work.


src/libcsg/modules/io/h5mdtrajectoryreader.cc
```diff
@@ -275,7 +274,7 @@ bool H5MDTrajectoryReader::NextFrame(Topology &top) {  // NOLINT const reference
     if (has_id_group_ != H5MDTrajectoryReader::NONE) {
       if (ids[at_idx] == -1)  // ignore values where id == -1
         continue;
-      atom_id = ids[at_idx] - 1;
```
JoshuaSBrown, line 278: @JensWehner, is this to make these files work with lammps or was this a bug?

JensWehner: I think that cam from master. I have not done that. I think

PR #397, opened 2019-08-14
Electrostatics update integration compare files lammps and dump file bounds

JoshuaSBrown:
Lammps comparison files updated to account for small numerical precision differences.


JoshuaSBrown:
@JensWehner we should probably see what @junghans says before merging this. Is it expected that these numerical values in these files would vary. The only thing that was changed was that the lammps file units were converted to real, and the lammps file readers and writers adjusted to account for this.

JensWehner:
agreed

JoshuaSBrown:
The conversion constants used for kj2kcal are in this pull request. https://github.com/votca/tools/pull/157

JensWehner:
Apparently, csg_force match kills a lot of significant digits. I approve this

PR #409, opened 2019-09-18
Fix includes

JensWehner:
Removes all using namespace from include headers


include/votca/csg/bead.h
```diff
@@ -58,24 +56,6 @@ class Bead : public BaseBead {
    */
   const int &getResnr() const { return residue_number_; }
 
-  /**
-   * get the mass of the bead
-   * \return bead mass
-   */
-  const double &getM() const {
```
JoshuaSBrown, line 65: Wouldn't it be better to mark this as deprecated? https://stackoverflow.com/questions/295120/c-mark-as-deprecated

src/libcsg/csgapplication.cc
```diff
@@ -34,31 +34,31 @@ void CsgApplication::Initialize() {
   TopologyReader::RegisterPlugins();
 
   if (NeedsTopology()) {
-    AddProgramOptions()("top", boost::program_options::value<string>(),
+    AddProgramOptions()("top", boost::program_options::value<std::string>(),
```
JoshuaSBrown, line 37: Again why not use 'using namespace std;' here in the source files?

src/libcsg/csgapplication.cc
```diff
@@ -393,21 +393,21 @@ CsgApplication::Worker::~Worker() {
...
+    ob->EndCG();
+  }
 }
 
 void CsgApplication::EvalConfiguration(Topology *top, Topology *top_ref) {
-  list<CGObserver *>::iterator iter;
-  for (iter = _observers.begin(); iter != _observers.end(); ++iter)
-    (*iter)->EvalConfiguration(top, top_ref);
+  for (CGObserver *ob : _observers) {
+    ob->EvalConfiguration(top, top_ref);
```
JoshuaSBrown, line 409: nice

JoshuaSBrown (dismissed):

src/libcsg/csgapplication.cc
JensWehner: as I said, I am not a big fan of using namespace and it is more explicit this way.

include/votca/csg/bead.h
JensWehner: It is not used anymore anywhere, as I changed the function calls. You could argue that some external programs could fail because of it.

junghans (approved):

PR #410, opened 2019-09-19
Refactor csg imc and fmatch

JensWehner:
A bit of cleanup should be merged after https://github.com/votca/csg/pull/409


src/tools/csg_fmatch.cc
```diff
@@ -468,60 +466,59 @@ void CGForceMatching::FmatchAccumulateData() {
     // FM residual is initially calculated in (kJ/(mol*nm))^2
     double fm_resid = residual.cwiseAbs2().sum();
 
-    // strange number is units conversion -> now (kcal/(mol*angstrom))^2
-    fm_resid /= 3 * _nbeads * _frame_counter * 1750.5856;
+    fm_resid /= 3 * _nbeads * _frame_counter;
```
junghans, line 469: What happened to the conversion factor?

JensWehner: is only used in the output 3 lines below and I added explicit units to the output

junghans: Great!

Issue #407, opened 2019-09-18
refactor imc solve


junghans:
Inside the scripts it calls:
```
linsolve.py --reg 0 CG-CG.packed CG-CG.packed.sol
```

which reads `CG-CG.packed.gmc` and `CG-CG.packed.imc` and writes results to `CG-CG.packed.sol`.

Example files from SPCE tutorials:
[CG-CG.packed.gmc](https://github.com/votca/csg/files/3623838/CG-CG.packed.gmc.txt)
[CG-CG.packed.imc](https://github.com/votca/csg/files/3623839/CG-CG.packed.imc.txt)
[CG-CG.packed.sol](https://github.com/votca/csg/files/3623840/CG-CG.packed.sol.txt)

junghans:
More realistic input:
[CG-CG.packed.gmc](https://github.com/votca/csg/files/3631584/CG-CG.packed.gmc.txt)
[CG-CG.packed.imc](https://github.com/votca/csg/files/3631585/CG-CG.packed.imc.txt)
[CG-CG.packed.sol](https://github.com/votca/csg/files/3631587/CG-CG.packed.sol.txt)

junghans:
Unpacked inputs:
[CG-CG.gmc](https://github.com/votca/csg/files/3636121/CG-CG.gmc.txt)
[CG-CG.idx](https://github.com/votca/csg/files/3636122/CG-CG.idx.txt)
[CG-CG.imc](https://github.com/votca/csg/files/3636124/CG-CG.imc.txt)

junghans:
[CG-CG.dpot.imc](https://github.com/votca/csg/files/3636922/CG-CG.dpot.imc.txt)
[CG-CG.gmc](https://github.com/votca/csg/files/3636924/CG-CG.gmc.txt)
[CG-CG.idx](https://github.com/votca/csg/files/3636926/CG-CG.idx.txt)
[CG-CG.imc](https://github.com/votca/csg/files/3636927/CG-CG.imc.txt)
[CG-CG.packed.gmc](https://github.com/votca/csg/files/3636929/CG-CG.packed.gmc.txt)
[CG-CG.packed.idx](https://github.com/votca/csg/files/3636931/CG-CG.packed.idx.txt)
[CG-CG.packed.imc](https://github.com/votca/csg/files/3636932/CG-CG.packed.imc.txt)
[CG-CG.packed.sol](https://github.com/votca/csg/files/3636933/CG-CG.packed.sol.txt)

[...]

PR #411, opened 2019-09-19
Csg imc solve

JensWehner:
implements https://github.com/votca/csg/issues/407

- [x] I do not know how to integrate this into the workflow now
 - [x] Some scripts will have to be deleted
 - [x] Merge after https://github.com/votca/csg/pull/410
 - [x] It is hard to test, because the data in the issue is numerically very unstable, which also explains why it likes to crash the tests.


junghans:
I will fix the scripting, but let me upload more realistic input data first.

JensWehner:
there was actually an error in my parser. So maybe the data is not so bad.

junghans:
More realistic data: https://github.com/votca/csg/issues/407#issuecomment-533143017

JensWehner:
I get exactly the same numbers, how can I write a test for this?

JensWehner:
would be an integration test with your data.

junghans:
@JensWehner see:
https://github.com/votca/csg/blob/master/src/tools/CMakeLists.txt#L115-L125
if you replace `csg_resample` by `csg_imc_solve` you are already half way there.

share/scripts/inverse/update_imc.sh
```diff
@@ -38,7 +37,7 @@ for group in $imc_groups; do
   # currently this is a hack! need to create combined array
   msg "solving linear equations for imc group '$group'"
   critical csg_imcrepack --in ${group} --out ${group}.packed
-  do_external imcsolver $solver  ${group}.packed ${group}.packed.sol $reg
+  critical csg_imc_solve --imcfile "${group}.packed.imc" --gmcfile "${group}.packed.gmc" --regularization "${reg}" --outputfile "${group}.packed.sol"
   critical csg_imcrepack --in ${group}.packed --unpack ${group}.packed.sol
```
junghans, line 41: @JensWehner, can be combine this series of `csg_imcrepack`, `csg_imc_solve`, `csg_imcrepack` into one thing?

junghans: I put the input files to `csg_imcrepack` here: https://github.com/votca/csg/issues/407#issuecomment-533553172

JensWehner: can you add the final output as well?

junghans: Without the `--unpack` option, it generates the `CG-CG.packed` files, with the `--unpack` option is generates the potentials e.g. *.imc.dpot`.

I uploaded a complete set here:
https://github.com/votca/csg/issues/407#issuecomment-533628560

JensWehner:
I have no really good idea, but I think considering that repack and csg_imc_solve do not work on the same things. I would do the merge of imc_repack and csg_imc_solve at a later stag.

junghans:
Sounds good to me.

JensWehner:
some scripts stuff still has to be removed correct?

junghans:
Done, forgot to push it.

junghans:
One last question, `csg_imc_solve` vs. `csg_imcsolve` (as in similar to `csg_imcrepack`), what do you like better as a name?

Also @DRosen285 wanted to test it, too.

JensWehner:
I like the one with two underscores a bit more, but consistency is also nice. So I have no strong opinion.

junghans:
@JensWehner ok then lets leave it like it is.

JensWehner:
@junghans you have to review and you can merge then from my side.

junghans:
Fine with me, let's give @DRosen285 another day to look at it.

DRosen285:
Sorry for the delay. 
Looks good to me. 
Just one comment following the discussion under #407 : it actually makes more sense to have one regularization parameter per group, instead per interaction. As the regularization term acts on the whole matrix, if multiple interactions are in the same group.

And one question just for curiosity: How do you determine the numerical stability of the IMC scheme? 
In the past I've detected it only based on how well the updated potentials perform in the next iteration. But it seems you keep track of the stability on the fly in csg_imc_solve.cc

JensWehner:
@DRosen285 I do not fully understand, but concerning the numerical instability. 

The problematic part of solving the problem is calculating the inverse of (A.T*A), because if the eigenvalues of (A.T*A) are very small you will nearly divide by zero. Small eigenvalues are normally a symptom of bad input data. So there is very little you can do. 

- Use regularisation, as you basicallly add a constant to each eigenvalue. 
- Calculate not the inverse of A.T*A, but the pseudo inverse. https://en.wikipedia.org/wiki/Generalized_inverse , basically you remove eigenvectors with very small eigenvalues (<epsilon) from the calculation. Depending on the epsilon, this will stabilize the numerics of your calculation.

DRosen285:
@JensWehner got it. I just got confused, because I used the pseudoinverse in combination with regularisation in the past. Because as I understand it, the pseudoinverse only helps to make a matrix invertible in case it is not. But one might still need regularization, because also for the pseudoinverse small eigenvalues might still exist (at least that is what I oberseved in the past).

junghans (approved):
LGTM

junghans:
I will fix #413 in a separate PR.

PR #416, opened 2019-09-24
cmake: turn votca_csg_boltzmann into an object lib

junghans:


JensWehner:
how come?

junghans:
The main reason was that, with this approach we don't have to compile the `csg_boltzmann` sources twice (once for the binary and once for the library).

JensWehner (dismissed):

PR #417, opened 2019-09-25
imc: make regularization group based

### LINKED ISSUE #413: imc: second regularization factor gets ignored
The problem is that `<reg>` is a per-interaction property, while for solver we loop over imc groups, which don't necessarily have to be the same.

We need to change <reg> to be a imc group property, e.g. looking in `cg.inverse.imc.regularization.<group_name>`.

junghans:
Fix #413


junghans:
Still need to run a couple of examples on this.

JensWehner (dismissed):
I am not really able to review this.

DRosen285:
> Still need to run a couple of examples on this.

The LJ example gets stuck. It seems there is a problem with the imc_repack. When I execute csg_imcrepack --in group_1 --out group_1 nothing happens.

Issue #418, opened 2019-09-26
imc tests hang on csg_repack


junghans:
If you run the IMC tutorials they hang in `csg_imcrepack` using https://github.com/votca/votca/pull/191:
```
cmake .. -DENABLE_TESTING=ON -DENABLE_REGRESSION_TESTING=ON -DENABLE_EXPERIMENTAL_TESTS=ON
make
ctest -V -R LJ
...
solving linear equations for imc group 'group_1' (regularization 1000)
```

`csg_imcrepack` was only touched in 16e43ee22a497122854ed7de2a6171528163b23b, if I revert that commit it fails on the unpack after running `csg_imc_solve` with:
```
Running critical command 'csg_imcrepack --in group_1.packed --unpack group_1.packed.sol'
terminate called after throwing an instance of 'std::runtime_error'
  what():  invalid range 61:60: begin, end and stride do not form a closed interval
```

junghans:
And this problems persists even before #411 (e.g. with ecad9be9c00bedc67b5145d15b515306365de641 and votca/csg-tutorials@9e75a8e237aa8a15a688545cdfd0c4b8f90d91f3)

However the spce/imc (`ctest -V -R spce_imc`) passes.

junghans:
Semi-good news, using  votca/votca#191 and reverting 16e43ee (`git revert 16e43ee`), I get the spce/imc to pass. So there are two bugs, one in 16e43ee and another one in the LJ example I still have to track down.

DRosen285:
The .packed.imc files look weird for the LJ example. It seems that single lines are getting written multiple times. In my case the same line gets written 60 times,.

junghans:
Another semi-good news, with `v1.5` the `LJ` example still works, the call sequence is something like:
```
Running critical command 'csg_imcrepack --in group_1 --out group_1.packed'
written group_1.packed.imc
written group_1.packed.gmc
written group_1.packed.idx
Running subscript 'solve_numpy.sh group_1.packed group_1.packed.sol 1000 1000 1000' (from tags imcsolver numpy) dir /home/christoph/votca/src/votca/csg/share/scripts/inverse
Running subscript 'linsolve.py --reg 1000 group_1.packed group_1.packed.sol' (from tags solve numpy) dir /home/christoph/votca/src/votca/csg/share/scripts/inverse
Running critical command 'sed -ie s/NaN/0.0/ group_1.packed.sol'
[...]

PR #419, opened 2019-09-26
Add more imc tests

junghans:
Might be useful for https://github.com/votca/csg/issues/418


src/tools/references/spce/CG-CG.gmc
```diff
@@ -0,0 +1,14 @@
...
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 
+-0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0 -0
```
JensWehner, line 11: sure that that is an empty matrix?

junghans: It is a one frame, so the fluctuation matrix is going to be zero.

But good point, we might want to slightly bigger system with more frames at some point.

junghans: Plus that was always the result, we just never check it ;-)

PR #420, opened 2019-09-26
Fixup IMC

### LINKED ISSUE #418: imc tests hang on csg_repack

junghans:
Fix #418


junghans:
```
      Start  95: Compare_csg_repack-unpack_output_1
10/27 Test  #95: Compare_csg_repack-unpack_output_1 .................   Passed    0.07 sec
      Start  96: Compare_csg_repack-unpack_output_2
11/27 Test  #96: Compare_csg_repack-unpack_output_2 .................   Passed    0.07 sec
      Start  97: Compare_csg_repack-unpack_output_3
12/27 Test  #97: Compare_csg_repack-unpack_output_3 .................   Passed    0.16 sec
```
which is good, but
```
	 91 - Compare_csg_repack_output_1 (Failed)
	 92 - Compare_csg_repack_output_2 (Failed)
	 93 - Compare_csg_repack_output_3 (Failed)
```
which is bad.

So packing is brocken, but  unpacking is fine.

DRosen285:
Don't know if this really helps, but when I comment out "imcio_write_matrix(name_out + ".gmc", gmc, &list);"
 packing does not hang anymore.
Nevertheless the packed.imc file that gets created still looks weird to me (I'm running the LJ example). So maybe there is the sth. going wrong with "imcio_write_dS" and "imcio_write_matrix".

JensWehner:
Yeah I probably messed the read and write up.

junghans:
@JensWehner it looks like the pack and unpack was done in order to workaround zeros in matrix, however due to the fact that we had to limit the iteration range to non-zero rdf regions for different reasons, this might not be needed anymore. I just tested this with all tutorials. Thanks to @DRosen285 for the hint and discussion on that.

Could you merge the unpack part (writing the .dpot.imc files) into csg_imc_solve ? And then we just can remove csg_imcrepack fully.

src/tools/csg_imc_solve.cc
```diff
@@ -104,4 +107,19 @@ void CG_IMC_solve::Run() {
   x.y() = -inverse * A.transpose() * B.y();
 
   x.Save(outputfile);
```
junghans: @DRosen285 is the output of the complete solution vector still useful if we also write the potentials?

src/tools/csg_imc_solve.cc
```diff
@@ -104,4 +107,19 @@ void CG_IMC_solve::Run() {
...
+
+  if (OptionsMap().count("idxfile")) {
+    std::string idxfile = _op_vm["idxfile"].as<std::string>();
+    std::vector<std::pair<std::string, votca::tools::RangeParser> > ranges =
+        votca::csg::imcio_read_index(idxfile);
+    votca::tools::Table tbl_in;
+    tbl_in.Load(outputfile);
```
junghans: @JensWehner maybe you can think about a better solution here than reading it in again.

src/tools/csg_imc_solve.cc
DRosen285: If with potentials you mean the potential update (dU or dpot) as well, then I think this can go away.

JensWehner:
What does it take to close this PR? I am a bit confused, I do not mind the C++ work, but I have no idea what is supposed to be going on? Can someone tell me, what is supposed to be done.

src/tools/csg_imc_solve.cc
JensWehner: Like this?

src/tools/csg_imc_solve.cc
junghans: Then we should move this, too!

src/tools/csg_imc_solve.cc
junghans: Superb!

src/tools/csg_imc_solve.cc
junghans: Done.

junghans:
@JensWehner I think now this is ready to merge.

DRosen285 (approved):

JensWehner:
if you guys like it then I am good with it

JensWehner (approved):

PR #426, opened 2019-10-03
cma.py: workaround for python3.8

junghans:
@tylerjereddy, it was actually an easy fix.

Related to https://github.com/votca/csg-tutorials/issues/4


tylerjereddy:
I see -- looks like the deprecation and removal were done because `time.clock()` was not portable while the newer thing is: https://docs.python.org/3/library/time.html#time.clock

JensWehner (approved):

PR #430, opened 2019-10-07
imc: drop crop potential

### LINKED ISSUE #429: IMC: remove potential crop
Related to https://github.com/votca/csg-tutorials/issues/59.

It seems to me that with only support non-zero rdf regions, we should be able to remove the potential crop in the IMC update.

/cc @DRosen285

junghans:
@DRosen285 can you check if this change makes a difference for your examples?

Fix #429


DRosen285 (dismissed):
It ran stable for 10 imc iterations.

DRosen285 (approved):

PR #432, opened 2019-10-17
overwrite

JensWehner:
added overwrite specifiers https://en.cppreference.com/w/cpp/language/override


junghans:
Duplicate of #436

PR #437, opened 2019-10-24
fixes for -Wextra -Wpedantic -Wshadow -Wconversion

junghans:
Part of https://github.com/votca/votca/pull/212

### DIFF
--- a/src/libcsg/modules/io/dlpolytopologyreader.cc
+++ b/src/libcsg/modules/io/dlpolytopologyreader.cc
@@ -219,7 +219,8 @@ bool DLPOLYTopologyReader::ReadTopology(string file, Topology &top) {
 #endif
 
       // read molecule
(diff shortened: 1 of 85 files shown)


src/libcsg/modules/io/dlpolytopologyreader.cc
```diff
@@ -219,7 +219,8 @@ bool DLPOLYTopologyReader::ReadTopology(string file, Topology &top) {
 #endif
 
       // read molecule
-      int id_map[natoms];
+      unique_ptr<long int[]> id_map =
+          unique_ptr<long int[]>{new long int[natoms]};
```
junghans: @JensWehner please check.

JensWehner: that sounds very complicated what we want is a `std::vector<int>(natoms);`

JensWehner (dismissed):
We can still rename 'long int' to `long`at some point. Otherwise thanks for the work

PR #439, opened 2019-10-31
remove bundled cma.py

### LINKED ISSUE #427: Use cma module provided by python
Once https://bodhi.fedoraproject.org/updates/FEDORA-2019-fcce755e99 goes in and Fedora 31 is our main testing plattform, we should use unbundle cma.py.

Related to #426, #185 and https://github.com/votca/csg-tutorials/issues/4

junghans:
Fix #427


junghans:
It fails now because of course Ubuntu has no python-cma module.

junghans:
> It fails now because of course Ubuntu has no python-cma module.

https://github.com/votca/csg-tutorials/pull/64

JensWehner:
and now?

PR #454, opened 2019-11-20
cmake: check for exact version of tools

### LINKED ISSUE #103: cmake: add version number in csg/tools check
```
csg builds only with exactly the same of tools, e.g.
1.2.0 with 1.2.0 and 1.3-dev with 1.3-dev.

We need to a an extra check for that to FindVOTCA_Tools.cmake
```

Original issue reported on code.google.com by `christop...@gmail.com` on 20 Jun 2011 at 8:41

junghans:
Fix #103


junghans:
```
 CMake Error at CMakeLists.txt:68 (find_package):
   find_package called with invalid argument "1.6-dev"
```
apparently `1.6-dev` doesn't count as a version. We could call `master` `9999`, but we still need to test what to do about release candidates, e.g. `1.5_rc1`.

JensWehner:
Hmmm no idea.

JensWehner:
What do other packages do?

junghans:
```
CMake Error at CMakeLists.txt:4 (find_package):
  find_package called with invalid argument "1.6rc1"
```
We might need to implement this in a different way.

junghans:
We should do the same for the other modules as well.

JensWehner:
yep,

PR #381, opened 2019-06-23
for/development/electrostatics joshs development

JoshuaSBrown:
This pull request must be merged at the same time as:
[tools](https://github.com/votca/tools/pull/154)
[csg](https://github.com/votca/csg/pull/381)
[xtp](https://github.com/votca/xtp/pull/275)
[csgapps](https://github.com/votca/csgapps/pull/38)

PR #457, opened 2019-11-22
Piece wize boundaries

JoshuaSBrown:
So this is a piece wise merge of my development branch https://github.com/votca/csg/pull/381, I'm going to break it up into a bunch of small merges because it is too much to do all at once. Here, the boundary class is updated.
Methods are made const
pragma once is included
header guards are updated
A clone method is introduced which allows safe copying of the class, which is needed because the boundarycondition class is stored as a dynamic pointer.


include/votca/csg/triclinicbox.h
```diff
@@ -29,12 +29,16 @@ class TriclinicBox : public BoundaryCondition {
   Eigen::Vector3d BCShortestConnection(
       const Eigen::Vector3d &r_i, const Eigen::Vector3d &r_j) const override;
 
-  eBoxtype getBoxType() override { return typeTriclinic; }
+  virtual std::unique_ptr<BoundaryCondition> Clone() const override {
```
JensWehner: Can we replace the clone method with a copy constructor and assignment operator

include/votca/csg/orthorhombicbox.h
```diff
@@ -29,12 +29,16 @@ class OrthorhombicBox : public BoundaryCondition {
   Eigen::Vector3d BCShortestConnection(
       const Eigen::Vector3d &r_i, const Eigen::Vector3d &r_j) const override;
 
-  eBoxtype getBoxType() override { return typeOrthorhombic; }
+  virtual std::unique_ptr<BoundaryCondition> Clone() const override {
```
JensWehner: we can actually make them `final`

JensWehner (commented):
I am not sure about the clone method. I think a proper copyconstructor would be more explicit and follow the intention.  Otherwise nice additions

include/votca/csg/triclinicbox.h
JoshuaSBrown: Sounds good

include/votca/csg/orthorhombicbox.h
JoshuaSBrown: yes good idea

include/votca/csg/triclinicbox.h
JensWehner: we need a move and move assignment as well then.

JoshuaSBrown: Actually, I was looking over this I'm not sure copy is appropriate, the reason for returning a unique_ptr to the base class allows you to dynamically store the child classes in a vector of base class pointers. I will keep it as is though if you think there is a better name for the method I'm open for ideas?

JensWehner: Why would you not have a copy constructor then?

JoshuaSBrown: Because you are not explicitly copying the object, the object is always being handled by pointers. Why bother adding code if you are just manipulating the pointers. The clone method is returning a unique pointer to the base class when it is called from a child class, even though it is copying the object that is not all it is doing, using a copy constructor would be misleading in this case.

JensWehner: okay I see your point.

junghans:
Misses code formatting.

JoshuaSBrown:
@JensWehner Can you review/approve?

junghans (approved):

PR #468, opened 2019-12-08
Piece wise beadmotif

JoshuaSBrown:
Changed from using pointer to reference
Added const noexcept keywords with appropriate methods
Added pragma once to header file 
Changed include from global path to relative


JensWehner (approved):
LGTM

PR #469, opened 2019-12-08
Piece wise beadlist

JoshuaSBrown:
Updated include guard
Removed unused include
Made includes relative where appropriate
Added const to method

### DIFF
--- a/include/votca/csg/beadlist.h
+++ b/include/votca/csg/beadlist.h
@@ -57,7 +57,7 @@ class BeadList {
   iterator begin() { return _beads.begin(); }
   iterator end() { return _beads.end(); }
 
-  Topology *getTopology() { return _topology; }
+  Topology *getTopology() const { return _topology; }
 
  private:
   std::vector<Bead *> _beads;
--- a/include/votca/csg/beadlist.h
+++ b/include/votca/csg/beadlist.h
@@ -57,7 +57,7 @@ class BeadList {
   iterator begin() { return _beads.begin(); }
   iterator end() { return _beads.end(); }
 
-  Topology *getTopology() const { return _topology; }
+  const Topology & getTopology() const { return _topology; }
 
  private:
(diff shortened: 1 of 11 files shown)


include/votca/csg/beadlist.h
```diff
@@ -56,7 +57,7 @@ class BeadList {
   iterator begin() { return _beads.begin(); }
   iterator end() { return _beads.end(); }
 
-  Topology *getTopology() { return _topology; }
+  Topology *getTopology() const { return _topology; }
```
JensWehner: want to return a pointer here or maybe a reference? And do we really want to return a "non const"?

JoshuaSBrown: This was a loaded request... :)

JensWehner (dismissed):
LGTM

src/libcsg/nblistgrid_3body.cc
```diff
@@ -78,7 +78,7 @@ void NBListGrid_3Body::Generate(BeadList &list1, BeadList &list2,
   }
 
   // check if both bead lists "have" the same topology
-  assert(list1.getTopology() == list2.getTopology());
+  assert(&(list1.getTopology()) == &(list2.getTopology()));
```
junghans, line 81: Why is this an `assert`?  Shouldn't if be an runtime error?

JensWehner: I do not know, my guideline is: 
If it can crash due to user input, it should be a runtime_error
If it can only crash due to wrong programming it should be an assert

junghans: `list1` and `list2` need to be in the same topology, so an assert makes sense here.

JoshuaSBrown:
@junghans @JensWehner can one of you approve?

junghans (approved):

PR #470, opened 2019-12-10
[WIP] fixed alignment problem

JensWehner:
certain eigen datastructures require alignment modifiers if stored in std:: datastructures, to avoid making the code more complicated, we revert back to dynamic sized matrices here.

PR #471, opened 2019-12-10
fixed alignment problem

JensWehner:
certain eigen datastructures require alignment modifiers if stored in std:: datastructures, to avoid making the code more complicated, we revert back to dynamic sized matrices here.

same as https://github.com/votca/csg/pull/470

Fix #470 

https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html

### DIFF
--- a/include/votca/csg/potentialfunctions/potentialfunctioncbspl.h
+++ b/include/votca/csg/potentialfunctions/potentialfunctioncbspl.h
@@ -67,7 +67,10 @@ class PotentialFunctionCBSPL : public PotentialFunction {
   double _dr;
   Eigen::VectorXd _rbreak;
 
-  // this shoudl not be a 4x4 matrix because of alignment problems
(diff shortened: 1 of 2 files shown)


include/votca/csg/potentialfunctions/potentialfunctioncbspl.h
```diff
@@ -67,7 +67,8 @@ class PotentialFunctionCBSPL : public PotentialFunction {
   double _dr;
   Eigen::VectorXd _rbreak;
 
-  Eigen::Matrix4d _M;
+  // this shoudl not be a 4x4 matrix because of alignment problems
```
junghans: Typo!

junghans (commented):
Could you explain the issue a bit better? The comment isn't very helpful.

junghans (approved):

JensWehner:
added a link above

PR #474, opened 2019-12-11
Piece wise beadstructure

JoshuaSBrown:
Corrected include guard
Added pragma once
Added const correctness to variables 
Added comments to method description
Simplified namespace TOOLS to tools
Switched from using pointer to references
Added const method accessor for getting bead

### DIFF
--- a/include/votca/csg/beadstructure.h
+++ b/include/votca/csg/beadstructure.h
@@ -26,8 +26,6 @@
 #include <votca/tools/graphalgorithm.h>
 #include <votca/tools/graphdistvisitor.h>
 
-namespace TOOLS = votca::tools;
-
 namespace votca {
 namespace csg {
 
@@ -77,12 +75,12 @@ class BeadStructure {
    *
    * The same bead cannot be added twice.
    **/
-  void AddBead(T *bead);
+  void AddBead(T &bead);
 
   /**
    * \brief Get the bead with the specified id
    **/
-  T *getBead(Index id);
+  T &getBead(Index id);
 
   /**
    * \brief Create a connection between two beads in the structure
@@ -97,7 +95,7 @@ class BeadStructure {
    **/
   std::vector<T *> getNeighBeads(Index index);
 
-  TOOLS::Graph getGraph();
+  tools::Graph getGraph();
   /**
    * \brief Compare the topology of two bead structures
    *
@@ -115,17 +113,17 @@ class BeadStructure {
  protected:
   void InitializeGraph_();
   void CalculateStructure_();
-  TOOLS::GraphNode BaseBeadToGraphNode_(T *basebead);
+  tools::GraphNode BaseBeadToGraphNode_(T *basebead);
 
   bool structureIdUpToDate = false;
   bool graphUpToDate = false;
   bool single_structureUpToDate_ = false;
   bool single_structure_ = false;
   std::string structure_id_ = "";
-  TOOLS::Graph graph_;
-  std::set<TOOLS::Edge> connections_;
+  tools::Graph graph_;
+  std::set<tools::Edge> connections_;
   std::unordered_map<Index, T *> beads_;
-  std::unordered_map<Index, TOOLS::GraphNode> graphnodes_;
+  std::unordered_map<Index, tools::GraphNode> graphnodes_;
 };
 
 /**********************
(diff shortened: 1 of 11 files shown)


include/votca/csg/beadstructure.h
```diff
@@ -76,27 +75,28 @@ class BeadStructure {
    *
    * The same bead cannot be added twice.
    **/
-  void AddBead(T *bead);
+  void AddBead(T &bead);
```
JensWehner: I am not truly happy with that interface, because it does not make the ownership clear. Who owns bead?

include/votca/csg/beadstructure.h
```diff
@@ -76,27 +75,28 @@ class BeadStructure {
...
 
   /**
    * \brief Get the bead with the specified id
    **/
-  T *getBead(Index id);
+  T &getBead(const Index &id);
```
JensWehner: nice

include/votca/csg/beadstructure.h
```diff
@@ -76,27 +75,28 @@ class BeadStructure {
...
-  void ConnectBeads(Index bead1_id, Index bead2_id);
+  void ConnectBeads(const Index &bead1_id, const Index &bead2_id);
 
   /**
    * \brief Return a vector of all the beads neighboring the index
    **/
   std::vector<T *> getNeighBeads(Index index);
 
-  TOOLS::Graph getGraph();
+  tools::Graph getGraph();
```
JensWehner, line 107: can it be const?

include/votca/csg/beadstructure.h
```diff
@@ -279,12 +281,19 @@ std::vector<T *> BeadStructure<T>::getNeighBeads(Index index) {
...
 template <class T>
-T *BeadStructure<T>::getBead(Index index) {
+T &BeadStructure<T>::getBead(const Index &index) {
+  assert(beads_.count(index));
+  return *beads_[index];
+}
+
+template <class T>
+const T &BeadStructure<T>::getBead(const Index &index) const {
   assert(beads_.count(index));
```
JensWehner: maybe want to throw a message with the assert

JensWehner (commented):
I do not know, for me the whole ownership in csg is unclear.

include/votca/csg/beadstructure.h
JoshuaSBrown: It is a copy there is no need to make const, and the copy is needs to be mutable so no, but good question.

include/votca/csg/beadstructure.h
JoshuaSBrown: What would you do instead? Shared pointer?

JoshuaSBrown: Use a copy?

include/votca/csg/beadstructure.h
JensWehner: I meant `Graph getGraph() const` not `const Graph getGraph()` sorry for the confusion

include/votca/csg/beadstructure.h
JensWehner: Does bead structure own beads or is it more a view on a collection of beads?

JensWehner: I do not know

include/votca/csg/beadstructure.h
JoshuaSBrown: No, unfortunately, it cannot be at this point. The graph is only generated internally when it is needed. When the getGraph method is called it checks to see if the graph is up to date, if it is it does not generate anything, but if it isn't the graph is generated. I could auto update the graph every time a bead is added but this would be expensive, particularly if the graph is never actually needed.

include/votca/csg/beadstructure.h
JoshuaSBrown: It was meant to act as a window, but I have changed it now so that it does not store pointers internally. This means one has to now be careful that if the beads change that they use a new beadstructure. Though only the masses and names of the beads are stored internally, and these are very unlikely to change.

include/votca/csg/beadstructure.h
JoshuaSBrown: I'll just get rid of the method.

JoshuaSBrown:
@JensWehner I hope this is more to your liking, I removed the internal pointers, so the beadstructure  no longer runs the risk of accessing invalid pointers.

include/votca/csg/beadstructure.h
JensWehner: window=view? Can you explain a bit more?

include/votca/csg/beadstructure.h
JensWehner: ahh okay

JoshuaSBrown:
@JensWehner hey does this satisfy your review? Can you approve?

JensWehner (approved):
Yeah it is definitely an improvement

PR #475, opened 2019-12-12
Piece wise beadmotifconnector

JoshuaSBrown:
Used pragma once
Const correctness applied
Improved testing
noexcept keyword added where appropriate
assertion test added in appropriate method


JoshuaSBrown:
@JensWehner hey can you take a look when you get the chance?

src/libcsg/beadmotifconnector.cc
```diff
@@ -15,21 +15,22 @@
  *
  */
 
-#include <votca/csg/beadmotifconnector.h>
+#include "../../include/votca/csg/beadmotifconnector.h"
```
junghans, line 18: Why?

JoshuaSBrown:
I believe it is a better practice. For instance if you build the project and it has previously installed it can link to the wrong files.

JoshuaSBrown:
It is safer to reserve global reference for the other repositories, and use the relative include paths internally, it doesn't look as nice though.

junghans:
I think it doesn't hurt, we add the local include path here anyhow:
https://github.com/votca/csg/blob/master/src/libcsg/CMakeLists.txt#L32

JoshuaSBrown:
My wording is probably a bit off, <> notation looks somewhere on the system, whereas "" will look in the current repo.

JoshuaSBrown:
I have had issues with this in the past but I am not sure if they were before or after those lines were placed in the cmake file.

junghans:
@JensWehner please review

JoshuaSBrown:
Ping @JensWehner

JensWehner (approved):

PR #483, opened 2019-12-20
Piece wise readerwriter units

JoshuaSBrown:
Added default units to all the file readers and writers, as well as default units for csg.


JensWehner (dismissed):
Can you explain a bit more what this does. please?

JoshuaSBrown:
@JensWehner this pull request does very little at the moment. I have added the default units of each of the file readers and writers, in coming pull requests I will be using the default units given in this pull request to convert between them using the unitconverter class in tools. That way you can read a pdb file in its default units and automatically convert it to a gromacs file in its default units.

include/votca/csg/pdbwriter.h
```diff
@@ -22,12 +22,15 @@
...
 #include <votca/csg/trajectorywriter.h>
 #include <votca/tools/constants.h>
+#include <votca/tools/unitconverter.h>
 
 namespace votca {
 namespace csg {
 
 class PDBWriter : public TrajectoryWriter {
  public:
+  const tools::DistanceUnit distance_unit = tools::DistanceUnit::angstroms;
```
junghans, line 32: Why `tools::` and not `csg::`?

junghans (approved):

include/votca/csg/pdbwriter.h
JoshuaSBrown: All of the types come from the unitconverter header in tools.

PR #514, opened 2020-04-17
Stable fixes

junghans:


JensWehner (commented):
why not always? seems a bit odd to not have it on 2 body interactions

junghans:
> why not always? seems a bit odd to not have it on 2 body interactions

The 3-body interactions are angles in radians and not box lengths, so comparing them makes no sense.

JensWehner:
okay I did not see the `!` in the if clause

PR #518, opened 2020-04-21
cma_processor.py: add support for cma-3

junghans:
cma-3 expect eps parameter to be a float, so added type=float. Also
replace obsolete optparse by argparse.


junghans:
Ok, this fixes the issue on Rawhide (cma-3) and doesn't break the build on other (cma-2).

PR #523, opened 2020-04-25
cmake: drop WITH_GMX in cmake in favor of auto-detect

junghans:
`WITH_GMX` was the last option left of that sort.


JensWehner (approved):
makes sense

PR #525, opened 2020-04-27
Add more csg_reupdate tests

### LINKED ISSUE #81: spce/re tutorials is broken in 1.6
```
Updating parameters
an error occurred:
Hessian NOT a positive definite!
This can be a result of poor initial guess or ill-suited CG potential settings or poor CG sampling.
```
reported here:
https://groups.google.com/d/msg/votca/wFNG1mmKwSU/9H4AZ_rfDQAJ

The tutorial still worked in v1.5.1

junghans:
Related https://github.com/votca/csg-tutorials/issues/81

I created a test, which succeeded in v1.5.1, but now fails in v1.6.

@JensWehner @JoshuaSBrown you touched `csg_reupdate.cc` the most please have a look.


JensWehner:
I cannot think of any change we had in those files. Hmm

JensWehner:
Have you printed out the hessian with both versions and saw if there was a change?

`cout<<_hs<<endl;`?

junghans:
v1.5.1:
```
37: Hessian
37: -2.84217e-14 -2.84217e-14 -5.68434e-14            0
37: -2.84217e-14 -1.13687e-13 -1.13687e-13            0
37: -5.68434e-14 -1.13687e-13 -2.27374e-13 -1.13687e-13
37:            0            0 -1.13687e-13 -4.54747e-13
37: End Hessian
```
stable (v1.6):
```
47: Hessian
47: 0 0 0 0
47: 0 0 0 0
47: 0 0 0 0
47: 0 0 0 0
47: End Hessian
```

the latter seems wrong.

junghans:
Here is how you can reproduce it:
```
cd votca
git checkout v1.5.1
git submodule update
git -C csg checkout debug_v1.5.1
cd build
cmake .. -DENABLE_TESTING=ON
make
ctest -V -R Run_csg_reupdate2
```
vs.
```
cd votca
git checkout stable
git submodule update
git -C csg checkout for/stable/csg_reupdate_fix
cd build
cmake .. -DENABLE_TESTING=ON
make
ctest -V -R Run_csg_reupdate2
```

src/tools/csg_reupdate.cc
```diff
@@ -618,10 +649,15 @@ void CsgREupdateWorker::EvalNonbonded(Topology *conf, PotentialInfo *potinfo) {
     votca::Index lamda_i = row - pos_start;
 
     double dU_i = 0.0;
+    cout<<"1 lamda_i " << lamda_i << endl;
     for (auto &pair_iter : *nb) {
       dU_i += potinfo->ucg->CalculateDF(lamda_i, pair_iter->dist());
```
junghans: @JensWehner I got closer, `CalculateDF()` is giving the wrong results.

junghans:
Good news, we didn't break csg_reupdate, it was just a nm vs AA conversion error. Bad news, I still have no idea why https://github.com/votca/csg-tutorials/issues/81 fails...

junghans:
Ok, boiled it down some more, run
```
csg_reupdate --top topol.tpr --trj traj.xtc --options settings.xml --hessian-check no
```
with these [inputs](https://github.com/votca/csg/files/4566207/input.zip) and the results (i.e. `CG-CG.pot.new`) will differ. However if you run the same with the `conf.gro` instead of `traj.xtc` the results are same. So my guess is now something is broken in the xtc trajectory reader.

junghans:
Same for the trr reader (with `--trj traj.trr` from `gmx trjcat -f traj.xtc -o traj.trr`), however running `csg_map --no-map --top topol.tpr --trj traj.xtc --out out.gro` and comparing `out.gro` with the output of `gmx dump -f traj.xtc` shows the position are read correctly.

junghans:
Ok, it differs using `out.gro`, while using `conf.gro` it doesn't, so the problem is when using more that one frame in the trajectory.

junghans:
`csg_stat` is fine, just `csg_reupdate` seems to be effected.

Here is the [out.gro](https://github.com/votca/csg/files/4566888/out.gro.txt) to trigger it.

JoshuaSBrown:
I can take a look tonight.

junghans:
@JoshuaSBrown adn @JensWehner, so make it a bit easier, here is the summary of the state again. If you run:
```
csg_reupdate --top topol.xml --trj traj.gro --options settings.xml --hessian-check no
```
with these [inputs](https://github.com/votca/csg/files/4575698/inputs.zip) (all text-based now) the resulting `CG-CG.pot.new` will differ between `stable` and `v1.5.1`.

You can build the specific versions of `csg_reupdate` with:
```
cd votca
git checkout <BRANCH>
git submodule update
cd build
cmake ..
make
```
where `BRANCH` is `stable` and `v1.5.1`.

JoshuaSBrown:
> v1.5.1:
> 
> ```
> 37: Hessian
> 37: -2.84217e-14 -2.84217e-14 -5.68434e-14            0
> 37: -2.84217e-14 -1.13687e-13 -1.13687e-13            0
> 37: -5.68434e-14 -1.13687e-13 -2.27374e-13 -1.13687e-13
> 37:            0            0 -1.13687e-13 -4.54747e-13
> 37: End Hessian
> ```
> 
> stable (v1.6):
> 
> ```
> 47: Hessian
> 47: 0 0 0 0
> 47: 0 0 0 0
> 47: 0 0 0 0
> 47: 0 0 0 0
> 47: End Hessian
> ```
> 
> the latter seems wrong.

are you sure? lol

JoshuaSBrown:
I found the problem. The old v1.5.1 is incorrect. In these lines:
This section of the old code in csg_reupdate is double counting the diagonal:
```
for (int row = 0; row < _nlamda; row++) {                                                                                                                                                                         
    _DS(row) += (-1.0 * _beta * _dUFrame(row));
    for (int col = row; col < _nlamda; col++){
        _HS(row, col) += ( _beta * _beta * _dUFrame(row) * _dUFrame(col));                                      
        _HS(col, row) += ( _beta * _beta * _dUFrame(row) * _dUFrame(col));
    }
}  
```
It should be:
```
for (int row = 0; row < _nlamda; row++) {                                                                                                                                                                         
    _DS(row) += (-1.0 * _beta * _dUFrame(row));
    for (int col = row; col < _nlamda; col++){
        _HS(row, col) += ( _beta * _beta * _dUFrame(row) * _dUFrame(col));                                      
        if(row!=col){
            _HS(col, row) += ( _beta * _beta * _dUFrame(row) * _dUFrame(col));
        }
    }
}  
```
version 1.6 gets around this because of what Jens did with Eigen:

```
_DS -= _beta * _dUFrame;
_HS += _beta * _beta * _dUFrame * _dUFrame.transpose();
```

JensWehner:
I do not know what is right or wrong there. I probably oversaw that problem. When you add the diagonal you get the old values back?

Good catch.

junghans:
@symashayak does the diagonal need to be double counted?

JoshuaSBrown:
I imagine the double counting is accidental not intentional, but ya, probably a good idea to find out for sure.

junghans:
> I imagine the double counting is accidental not intentional, but ya, probably a good idea to find out for sure.

@symashayak is the expert here, there are some stabilization scheme, where adding the trace makes inversion more stable. Plus in the current implementation the physics is wrong.

junghans:
> I do not know what is right or wrong there. I probably oversaw that problem. When you add the diagonal you get the old values back?

I can at least confirm that adding the diagonal back in gives the v1.5.1 result back.

/cc @DRosen285

JoshuaSBrown:
Ping @symashayak

DRosen285:
I did some testing with docker versions 1.4, 1.5.1 and stable.
The spce/re tutorial worked with 1.4 and 1.5.1 and yield comparable results.
In stable the tutorial failed.
What I noticed, in v1.4 the diagonal gets counted only once 
https://github.com/votca/csg/blob/bb041498f0f8ef85c7f5849a6c686eb420fe6149/src/tools/csg_reupdate.cc#L567-L575

So the double  counting @JoshuaSBrown noticed seems to got introduced when going from version 1.4 to 1.5. 

@JensWehner any thoughts?

junghans:
09e21c14e65dea82616814019a40a29bd70cdded is where @JensWehner added the double counting.

JensWehner:
So it is right now, but gives the wrong results? I am sorry.

@DRosen285 thanks.

How come that the tests did catch it?

JoshuaSBrown:
Great, so this means we have an error somewhere else?

junghans:
> So it is right now, but gives the wrong results? I am sorry.

No, it seems we broke it between v1.4 and v1.5 already when moving to Eigen. v1.4 correct results, v1.5 reasonable (but maybe not fully correct) results, v1.6 wrong results.

junghans:
> How come that the tests did catch it?

We added the test in v1.5, so the reference data is wrong as well.

junghans:
@jens do you remember why you added 09e21c1?

JensWehner:
The reason was boost offered a symmetric matrix type, Eigen does not really have that so i had to add the other elements as well, I did not think about the diagonal though.

junghans:
Could you try to fix the diagonal part?

DRosen285:
Bad news: The tutorial is still broken. There must be sth. going wrong additionally compared to v1.4.

symashayak:
I reviewed the csg_reupdate.cc, other than the diagonal double counting, which seems to have been fixed, I don't see any issue there.

junghans:
Well the hessian is still not positive definite. Any idea why the tutorial doesn’t work?

DRosen285:
The tutorial works. I accidentally used the wrong version of csg_reupdate after recompiling it in the docker container. I'm sorry.

JensWehner:
so it works now?

junghans:
> Well the hessian is still not positive definite. Any idea why the tutorial doesn’t work?

Ok, I take it back, the tutorial seems to work again! I shouldn't have listened to @DRosen285 ;-)

> so it works now?

It seems so! @DRosen285 could you please have a look if the results make sense?

DRosen285:
Results are different compared to v1.4, but they still look reasonable to me.

PR #528, opened 2020-05-06
CI: sync changes from votca/votca

junghans:


JensWehner (approved):

junghans:
I think we should remove the gmx 2016 builds, but I will do that in stable.

PR #531, opened 2020-05-15
Standardize the include format

JoshuaSBrown:
[In response to this issue](https://github.com/votca/votca/issues/300)


src/libcsg/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ configure_file(votca_config.h.in ${CMAKE_CURRENT_BINARY_DIR}/votca_config.h)
 
 #CMAKE_CURRENT_BINARY_DIR for gitversion.h /votca_config.h
 target_include_directories(votca_csg PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
+  $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/votca/csg>
```
junghans: I think we don.'t need that line.

src/libcsg/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ configure_file(votca_config.h.in ${CMAKE_CURRENT_BINARY_DIR}/votca_config.h)
...
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/votca/csg>
```
junghans: And this is only need if we want to use the public header without `votca/csg` prefix in our tools. Do we want that?

junghans (changes requested):

src/libcsg/CMakeLists.txt
junghans: I think you did a look of `#include "../../votca/csg/foo.h`, so that should be needed.

JoshuaSBrown: No we definitely want the full prefix.

JoshuaSBrown: Why do we want these lines in xtp then?

junghans: I made a comment in xtp, the one line is unnecessary there, too.

The other one is needed as `votca_config.h` is a public header in xtp, but in csg it is private. (Not sure why it needs to be public in xtp)

junghans: I made a comment in xtp, the one line is unnecessary there, too.

The other one is needed as `votca_config.h` is a public header in xtp, but in csg it is private. (Not sure why it needs to be public in xtp)

JoshuaSBrown: ooh, ok so I do need this line because of the config file:

```
 $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/votca/csg>
```

junghans (approved):

src/libcsg/CMakeLists.txt
junghans: In xtp, but not csg.

junghans: For the xtp votca_config.h only, xtp doesn't know about csg's votca_config.h

JensWehner (approved):

PR #534, opened 2020-05-19
cmake: rename REGRESSIONTEST_TOLERANCE to INTEGRATIONTEST_TOLERANCE

junghans:
I never liked the name of this option and now it conflicts with https://github.com/votca/csg-tutorials/pull/88, so renaming.


JensWehner (approved):

JensWehner:
Like it is a better name

PR #539, opened 2020-05-28
correct search for gmx executable

JensWehner:


junghans:
Should that go into stable?

JensWehner:
do not care. What do you think?

junghans (approved):

PR #542, opened 2020-05-30
include formatting in lib

JoshuaSBrown:


junghans:
The problem is that gromacs as well as eigen define `real`.

src/libcsg/beadmotifalgorithms.cc
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009-2020 The VOTCA Development Team (http://www.votca.org)
+ * Copyright 2009-2019 The VOTCA Development Team (http://www.votca.org)
```
junghans: Huh?

src/libcsg/cgmoleculedef.cc
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009-2020 The VOTCA Development Team (http://www.votca.org)
+ * Copyright 2009-2019 The VOTCA Development Team (http://www.votca.org)
```
junghans: Huh?

src/libcsg/beadmotifalgorithms.cc
JoshuaSBrown: That's weird. I might have a old version of the git pre-commit hook in that repo.

JoshuaSBrown:
> The problem is that gromacs as well as eigen define `real`.

I figured that, that was the problem. I'm wondering how this should be handled, it is really annoying that the namespace scope is not being properly handled. What if we put all files from gromacs in their own namespace would this fix the problem.

```
namespace gromacs {
#include <gromacs...etc> 
}
```

I guess the real issue is that we are including a private header file to read in their .tpr files, which is dangerous and has lead to the problems we are currently experiencing. I could contact them and see if they would be willing to support a file reader in their actual api which is correctly scoped.

junghans:
> > The problem is that gromacs as well as eigen define `real`.
> 
> I figured that, that was the problem. I'm wondering how this should be handled, it is really annoying that the namespace scope is not being properly handled. What if we put all files from gromacs in their own namespace would this fix the problem.
> 
> ```
> namespace gromacs {
> #include <gromacs...etc> 
> }
> ```
> 
> I guess the real issue is that we are including a private header file to read in their .tpr files, which is dangerous and has lead to the problems we are currently experiencing. I could contact them and see if they would be willing to support a file reader in their actual api which is correctly scoped.

Sure, go for it, maybe not as part of this PR.

junghans (changes requested):
Add a changelog please.

junghans (approved):

Issue #535, opened 2020-05-21
H5MD reader: skip the first frame


junghans:
If I run:
```
csg_stat --nt 1 --options settings.xml --top topol.xml --trj traj.h5 --begin 0 --first-frame 0
```
with these [inputs](https://github.com/votca/csg/files/4664174/input.zip), the output (`CG-CG.dist.new`) is all zeros, even though the positions in `traj.h5` looks ok.

Longer trajectories works fine as well.

junghans:
In contrast:
```
csg_stat --nt 1 --options settings.xml --top topol.xml --trj spce.gro --begin 0 --first-frame 0
```
works.

jkrajniak:
I have debug a bit the example and the simulation script (so nice it is included in h5md file). I think the zeros are not because of not reading the first frame but because the .h5 uses angstroms and the settings.xml are in nm? If you change `<max>` to 9.0 then it calculates distribution.

junghans:
Good catch it seems we forgot to fix the reader after merging https://github.com/votca/csg-tutorials/pull/34. The simulation is now done in Angstrom, but the reader still thinks numbers are in nm.

Is there a field to store the units in the h5md format?

jkrajniak:
There is a module called `units` which holds the information about the unit system (https://nongnu.org/h5md/modules/units.html) This is nothing more than an `unit` attribute assigned to the dataset. However, if I am correct Espresso does not use this attribute to store the units (which I admit, could be beneficial)

junghans:
VOTCA internal units are nm, hence the missing factor 10.

As a workaround we could assume h5md is in Angstrom if no unit is given.

@JoshuaSBrown what do you think?

@jkrajniak is there are way to set the unit manually in the espresso script?

jkrajniak:
> As a workaround we could assume h5md is in Angstrom if no unit is given.

If VOTCA expects that length is expressed in nm then I would assume that the numbers in
H5MD are in nm units (so no conversion) and do the conversion only if the `units` is enabled. Just to follow the principle of garbage-in-garbage-out.
> 
> @JoshuaSBrown what do you think?
> 
> @jkrajniak is there are way to set the unit manually in the espresso script?

AFAIK not without modifying the `h5md.H5md` module in Espresso

JoshuaSBrown:
I think if there is a units attribute in h5md it should be used, and if it is not specified either a warning should be thrown indicating that a default is being used or an error should be thrown.

[...]

PR #549, opened 2020-06-14
Read units stored in H5MD file

jkrajniak:
The extension reads units stored in H5MD attributes and do the conversion.


src/libcsg/modules/io/h5mdtrajectoryreader.h
```diff
@@ -180,6 +183,24 @@ class H5MDTrajectoryReader : public TrajectoryReader {
...
   int vec_components_;
 
+  // Length scaling - VOTCA internally uses nm as the length unit.
+  bool unit_module_enabled_ = false;
+  double length_scaling_ = 1.0;
+  double velocity_scaling_ = 1.0;
+  double force_scaling_ = 1.0;
+
+  // Convert map from SI prefixes to nm (VOTCA base length unit).
+  std::unordered_map<std::string, double> length_units_votca_scaling_factors = std::unordered_map<std::string, double>(
```
JensWehner: I think an unordered map for 5 element is a bit overkill, but I do not care in the end.

src/libcsg/modules/io/h5mdtrajectoryreader.cc
```diff
@@ -22,6 +22,7 @@
 
 // Third party includes
 #include <hdf5.h>
+#include <boost/regex.hpp>
```
JensWehner, line 26: I think std:: has a regex as well now? I am not sure if we still use it, because of some distribution problem @junghans

junghans: I don't think boost-regex is big burden.

junghans:
A unit test would be nice, but as we don't have a h5md writer and I don't want to store binary files in the repo, there isn't much we can do right now.

src/libcsg/modules/io/h5mdtrajectoryreader.cc
```diff
@@ -345,5 +367,46 @@ void H5MDTrajectoryReader::ReadBox(hid_t ds, hid_t ds_data_type, Index row,
...
+      auto votca_scaling_factor = votca_units_scaling_factors.find(v);
+      if (votca_scaling_factor != votca_units_scaling_factors.end()) {
+        scaling_factor =
+            scaling_factor * pow(votca_scaling_factor->second, unit_pow);
+      }
+    }
+    std::cout << "Found units " << value << " for " << unit_type << ".";
+    if (scaling_factor != 1.0) {
+      std::cout << " Values scaled by a factor " << scaling_factor;
+    }
```
JoshuaSBrown, line 404: This is good!

src/libcsg/modules/io/h5mdtrajectoryreader.h
```diff
@@ -180,6 +184,22 @@ class H5MDTrajectoryReader : public TrajectoryReader {
...
+  double force_scaling_ = 1.0;
+
+  // Convert map from SI prefixes to nm (VOTCA base length unit).
+  std::unordered_map<std::string, double> votca_units_scaling_factors =
+      std::unordered_map<std::string, double>({{"fm", 0.000001},
+                                               {"pm", 0.001},
+                                               {"nm", 1.0},
+                                               {"A", 0.1},
+                                               {"um", 1000.0}});
+
```
JoshuaSBrown, line 200: Is there anyway we can use the unit converter in tools instead?

JoshuaSBrown (approved):
Thanks for doing this. My comments are optional, the unit converter in tools could be adjusted to help with the unit conversions but I will leave it to you as to whether you want to take the time to do that. 

Please update the changelog though.

jkrajniak:
> Thanks for doing this. My comments are optional, the unit converter in tools could be adjusted to help with the unit conversions but I will leave it to you as to whether you want to take the time to do that.
> 

I haven't put this in tools as the code here is strictly related to the H5MD reader.

junghans:
@votca-bot changelog: read units stored in H5MD file

junghans:
Writing of units was added to espresso in https://github.com/espressomd/espresso/pull/3751, reading is added here, to fix https://github.com/votca/csg/issues/535 we just need update our tutorial.

junghans (approved):

JensWehner (approved):

Issue #543, opened 2020-05-30
GROMACS File Readers


JoshuaSBrown:
@JensWehner you had trouble with gromacs file readers before, did you ever get the problem fixed. I'm getting tones of linker errors now as well.

JoshuaSBrown:
I'm wondering if all the gmx files now need a refactor.

junghans:
>  I'm getting tones of linker errors now as well.

Any details?

JensWehner:
I had an old version of a gromacs libarray around, which caused those errors. Everything is fine now, it was a user error.

JoshuaSBrown:
We might need to update cmake,  

Found GROMACS: /usr/lib/x86_64-linux-gnu/libgromacs_d.so (found suitable version "2019.6", minimum required is "2018")

Here is a list of the errors 

../libcsg/libvotca_csg.so.7: undefined reference to `gmx_mtop_t::gmx_mtop_t()'
../libcsg/libvotca_csg.so.7: undefined reference to `read_first_frame(gmx_output_env_t const*, t_trxstatus**, char const*, t_trxframe*, int)'
../libcsg/libvotca_csg.so.7: undefined reference to `open_trx(char const*, char const*)'
../libcsg/libvotca_csg.so.7: undefined reference to `close_trx(t_trxstatus*)'
../libcsg/libvotca_csg.so.7: undefined reference to `output_env_init(gmx_output_env_t**, gmx::IProgramContext const&, time_unit_t, bool, xvg_format_t, int)'
../libcsg/libvotca_csg.so.7: undefined reference to `read_next_frame(gmx_output_env_t const*.,. /t_trxstatuslibcsg*/,libvotca_csg.so.7 :t_trxframe *undefined) 'reference
 .to. /`libcsggmx_mtop_t/:libvotca_csg.so.7::gmx_mtop_t (undefined) 'reference
 .to. /`libcsggmx_mtop_t/:libvotca_csg.so.7::~ gmx_mtop_tundefined( )reference'
to. .`/read_first_framelibcsg(/gmx_output_env_tlibvotca_csg.so.7 :const *undefined,  referencet_trxstatus *to* ,` write_trxframechar( t_trxstatusconst**,,  t_trxframet_trxframe**,,  gmx_conect_tint*))''

.collect2: error: ld returned 1 exit status
./libcsg/libvotca_csg.so.7: undefined reference to `open_trx(char const*, char const*)'
[...]

PR #554, opened 2020-06-21
Updated find gromacs to work with spack installations

JoshuaSBrown:
Trying to resolve issues with using spack installation of gromacs with VOTCA. https://github.com/votca/csg/issues/543

The largest problem appears to be that the location of the version.h file is in a different folder gmxapi, and by default will install with mpi added to the gmx library and executable names.


junghans:
Sorry, I don't get the point of this PR, in spack `votca-csg` forces gromacs to be compiled without mpi: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/votca-csg/package.py#L35
and checking the version.h header in api folder might work at cmake time, but to actually compile the gromacs file reader, gromacs needs to be build with `-DGMX_INSTALL_LEGACY_API=ON`, which spack does:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/gromacs/package.py#L96

That all said, additionally `votca-csg` depends on `gromacs@:2019.9999` meaning spack won't pull any gromacs version newer than 2019.* as it would break csg's iterative functionalities.

Compiling csg against an mpi-enabled version of gromacs has many unwanted side effects and I am not sure if it is worth our time. Gromacs uses mpi only in the mdrun binary and hence newer versions of gromacs are recommended to be compiled with `-DGMX_BUILD_MDRUN_ONLY=ON` for mpi, which doesn't generate a libgromacs_mpi anymore. And this is basically what e.g. Fedora does, compiling gromacs twice once without mpi for `libgromacs` and once with mpi for `mdrun` only.

JoshuaSBrown:
Wow, ok i didn’t know all of that. All I wanted to be able to do was use spack to handle the gromacs installs cause I kept running into linking issues using the system installs. I didn’t realize there were going to be issues introduced by linking with gromacs_mpi. Do you know how to only install gromacs without mpi using spack? Maybe this will solve the problem I’m having.

junghans:
You can just do `spack install --only dependencies votca-csg@master` or if you just want gromacs without mpi, `spack install gromacs~mpi`.

PR #572, opened 2020-08-03
refactored some tests to use files instead of source code data

JensWehner:
Refactored unit tests using more files and less stuff in the sourcecode.


junghans (approved):

JensWehner:
@votca-bot changelog: refactored tests to use files instead of data in source code

Issue #119, opened 2015-08-18
Code review request


GoogleCodeExporter:
```
please confirm
```

Original issue reported on code.google.com by `konstantin.koschke` on 27 Oct 2011 at 10:41

PR #575, opened 2020-08-10
maxindex correction in csg_resample

marvinbernhardt:
This might seem like a cosmetic change. But when I use target (dihedral) distributions calculated with csg_stat and then use csg_resample (called from csg_inverse) with the same settings.xml then I loose one data point.

I'm not sure if this could break something somewhere else, though.


src/tools/csg_resample.cc
```diff
@@ -171,7 +171,7 @@ int main(int argc, char **argv) {
           if (in.x(i) < sp_min) {
             minindex = i;
           }
-          if (in.x(i) < sp_max) {
+          if (in.x(i) <= sp_max) {
```
junghans, line 174: `=` comparison for float is usually not a good idea. @JensWehner what would you do here?

junghans:
Seems to break a test:
```
50/86 Test #119: integration_Compare_csg_resample_cubicfit_output ...***Failed    0.04 sec
Comparing table_cubicfit and /__w/csg/csg/votca/csg/src/tools/references/csg_resample/table_cubicfit to a tolerance of 5e-05
Diff:
row 0 col 1:	0.9324530711 vs 0.9739102263
row 1 col 1:	2.21049713 vs 2.145689133
row 2 col 1:	2.735043252 vs 2.676548033
row 3 col 1:	2.099478317 vs 2.19769654
row 4 col 1:	1.284728477 vs 1.428862755
row 5 col 1:	1.281849978 vs 1.167379552
row 6 col 1:	1.734880172 vs 1.432480399
row 7 col 1:	2.079710618 vs 2.058572883
row 8 col 1:	2.26666861 vs 2.918860478
row 9 col 1:	2.374690376 vs 3.896245629
```

marvinbernhardt:
> Seems to break a test:
> 
> ```
> 50/86 Test #119: integration_Compare_csg_resample_cubicfit_output ...***Failed    0.04 sec
> Comparing table_cubicfit and /__w/csg/csg/votca/csg/src/tools/references/csg_resample/table_cubicfit to a tolerance of 5e-05
> Diff:
> row 0 col 1:	0.9324530711 vs 0.9739102263
> row 1 col 1:	2.21049713 vs 2.145689133
> row 2 col 1:	2.735043252 vs 2.676548033
> row 3 col 1:	2.099478317 vs 2.19769654
> row 4 col 1:	1.284728477 vs 1.428862755
> row 5 col 1:	1.281849978 vs 1.167379552
> row 6 col 1:	1.734880172 vs 1.432480399
> row 7 col 1:	2.079710618 vs 2.058572883
> row 8 col 1:	2.26666861 vs 2.918860478
> row 9 col 1:	2.374690376 vs 3.896245629
> ```

To me the new values seem more consistent. The input table is 
```
0 1
1 2
2 3
3 2
4 1
5 2
6 1
7 2
8 3 
9 2
```
The command is `csg_resample --in table_in --out table_cubicfit --type cubic --grid 0:1:9 --fitgrid 0:2:9`. I have not investigated this further, but it seems that the old cubicfit was ignoring the value at x=9.

junghans:
@votca-bot changelog: corrected max index in csg_resample

src/tools/csg_resample.cc
```diff
@@ -171,7 +171,7 @@ int main(int argc, char **argv) {
           if (in.x(i) < sp_min) {
```
junghans, line 171: Do we need to do the same for `minindex`?

marvinbernhardt: Really not sure about it..

src/tools/csg_resample.cc
JensWehner: Yes I am not really sure. In real examples it should not matter so much and we rely a bit on a corner case here. I am not sure atm

JensWehner: I think the major question is what should a range 1:10 mean is it incluive or exclusive.

junghans: @JensWehner  How about `in.x(i) < sp_max * (1 + DBL_EPSILON)`? (see http://www.cplusplus.com/reference/cfloat/).

junghans: > I think the major question is what should a range 1:10 mean is it incluive or exclusive.

Inclusive.

JensWehner: if we do it like this I would use https://en.cppreference.com/w/cpp/types/numeric_limits instead of the c way. i.e. `std::numerics<Index>::epsilon()`. Why not `sp_max+std::numerics<Index>::epsilon()`, although I think that would be equivalent to `in.x(i) <= sp_max`

junghans (approved):

PR #576, opened 2020-08-10
periodic dihedral extrapolation

marvinbernhardt:
When dealing with dihedrals I ran into problems with the extrapolation. potential_to_gromacs.sh would extrapolate both sides to 180°. If the new potential is then interpolated and the force is calculated, one obtains a very high force for x=180. 

This pull request is the solution I came up with. It will ignore the gradient of the last points on the right side and extrapolate linearly to the same value one finds on the left side.


junghans:
So we set the force to same value on both side (https://github.com/votca/csg/blob/master/share/scripts/inverse/table_to_xvg.pl#L101)
but the value is too high, correct?

share/scripts/inverse/potential_extrapolate.sh
```diff
@@ -90,9 +90,12 @@ intermediate="$(critical mktemp "${input}.onlyleft.XXXXX")"
 if [[ $pot_type = "non-bonded"  ]]; then
   do_external table extrapolate --function ${lfct:-exponential} --avgpoints $avg_points --region left "${input}" "${intermediate}"
   do_external table extrapolate --function ${rfct:-constant} --avgpoints 1 --region right "${intermediate}" "${output}"
-elif [[ $pot_type = "bond"  || $pot_type = "angle" || $pot_type = "dihedral" ]]; then
+elif [[ $pot_type = "bond"  || $pot_type = "angle" ]]; then
   do_external table extrapolate --function ${lfct:-linear} --avgpoints $avg_points --region left "${input}" "${intermediate}"
   do_external table extrapolate --function ${rfct:-linear} --avgpoints $avg_points --region right "${intermediate}" "${output}"
+elif [[ $pot_type = "dihedral" ]]; then
+  do_external table extrapolate --function ${lfct:-linear} --avgpoints $avg_points --region left "${input}" "${intermediate}"
```
junghans: Why would you not use `periodic` on the left side as well?

share/scripts/inverse/potential_extrapolate.sh
```diff
@@ -90,9 +90,12 @@ intermediate="$(critical mktemp "${input}.onlyleft.XXXXX")"
...
+  do_external table extrapolate --function ${rfct:-periodic} --avgpoints $avg_points --region right "${intermediate}" "${output}"
```
junghans: You will need to fix the help above in this script as well.

junghans:
@marvinbernhardt I am all for merging this as dihedrals aren't very well supported currently. Have a look at my comments above. Do dihedrals always need to go from 0 to 180?

junghans:
@votca-bot changelog: add periodic extrapolation for dihedrals

marvinbernhardt:
> So we set the force to same value on both side (https://github.com/votca/csg/blob/master/share/scripts/inverse/table_to_xvg.pl#L101)
> but the value is too high, correct?

Yes, precisely

share/scripts/inverse/potential_extrapolate.sh
marvinbernhardt: So the scheme I have here is to extrapolate the left side normally and then extrapolate the right side to meet leftmost value. I think this scheme should suffice in most cases. A more elaborate scheme would be to pre-extrapolate on both sides, save the value at the boundaries, and extrapolate both sides to their average. But that would require a restructuring of the whole script and possibly others.

marvinbernhardt:
Ups, sorry reverted those codacy changes. Will change again.

junghans:
@marvinbernhardt I undid the last two commits, was there a special reason to split the option parsing? I think we can alllow periodic for the left as well, even though we don't have a use case yet.

junghans (approved):

PR #578, opened 2020-08-16
fix multicomponent csg_stat_imc

marvinbernhardt:
fix https://github.com/votca/csg/issues/577

PR #579, opened 2020-08-16
fix multicomponent csg_stat_imc

### LINKED ISSUE #577: IMC multicomponent working?
Is there a working implementation of multicomponent IMC?

I'm trying with csg-tutorials/propane/imc, but only A-A gets updated. I took a look in src/tools/csg_stat_imc.cc, but with my low C++ skills I can't tell if multicomponent sampling of the IMC matrix is implemented or not. The resulting matrix all.gmc is zero everywhere except for the A-A block.

junghans:
fix #577 

replacement for #578


junghans:
@votca-bot changelog: fix multicomponent imc

PR #581, opened 2020-08-16
cmake: add multicomp. test imc test

junghans:


junghans:
Not quite happy with this test, as it is more than a MB in size for the data files.

@DRosen285 could you 

- make a system setup with less particles (the `conf.gro` is big)
- that uses normal LJ (functional) interactions instead of tables (tables are half of the size)
- upload a zip of the `conf.gro`, `grompp.mdp` and `topol.top` here
?

DRosen285:
> Not quite happy with this test, as it is more than a MB in size for the data files.
> 
> @DRosen285 could you
> 
>     * make a system setup with less particles (the `conf.gro` is big)
> 
>     * that uses normal LJ (functional) interactions instead of tables (tables are half of the size)
> 
>     * upload a zip of the `conf.gro`, `grompp.mdp` and `topol.top` here
>       ?
[smaller_system.zip](https://github.com/votca/csg/files/5081529/smaller_system.zip)

junghans:
Thanks, will get onto that soon!

junghans:
It seems running mdrun is too unstable let me have a look if we can get away with a two frame trajectory.

junghans:
Hmm, with 2 frames, I get output files full of `-0`.

junghans:
Let see if it passes now, then I will rebase everything into a single commit.

junghans:
@JensWehner please review.

@DRosen285 Thanks for the inputs, I used your setup, but I put the traj in there instead of running mdrun.

JensWehner (approved):
Our tests are getting longer and longer. I am happy with it. Good work

PR #597, opened 2020-09-16
Add FindEigen3.cmake

jkrajniak:
I have encountered problem with installing latest csg on HPC. Mainly `cmake` ended with the following message
```
CMake Error at CMakeLists.txt:68 (find_package):
  Could not find a package configuration file provided by "Eigen3" (requested
  version 3.3.0) with any of the following names:

    Eigen3Config.cmake
    eigen3-config.cmake

  Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
  "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
  provides a separate development package or SDK, be sure it has been
  installed.
```

I have checked the installed Eigen3 library and in fact these two files weren't there (.cmake). To solve it
I have copied FindEigen3.cmake from https://gitlab.com/libeigen/eigen/-/blob/master/cmake/FindEigen3.cmake
and this solved the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ else (BUILD_MANPAGES)
 endif (BUILD_MANPAGES)
 
 find_package(Boost 1.53.0 REQUIRED COMPONENTS program_options filesystem system regex)
-find_package(Eigen3 3.3.0 NO_MODULE REQUIRED)
+find_package(Eigen3 3.3.0 REQUIRED)
 find_package(Threads REQUIRED)
 if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   set_package_properties(Threads PROPERTIES TYPE REQUIRED PURPOSE "Used for thread parallelization")
--- a/CMakeModules/FindEigen3.cmake
+++ b/CMakeModules/FindEigen3.cmake
@@ -1,0 +1,110 @@
+# - Try to find Eigen3 lib
+#
+# This module supports requiring a minimum version, e.g. you can do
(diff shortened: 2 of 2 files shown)


junghans:
I am undecided, on the one hand I think the eigen3 installation on your cluster is just broken. On the other hand his fix is simple enough.

The real downside of adding custom find modules is that the get out of sync all the time and sometimes even become incompatible to previous versions.

Maybe we should check the config first and only afterwards use the module to find it.

JensWehner:
It should probably also end up in the tools module and not in csg?

JensWehner:
but it sounds a lot like a broken eigen installation

junghans:
There is another issue: when using a find module instead of a config, one has to install it as well, because we run `find_package(Eigen3)` in our exported target as well. I really not in favor of going down the route of custom modules again.

Currently we have two non-trivial library modules left `fftw3` and `mkl`, the former we could detect with pkg-config using its imported target that latter will hopefully be fixed at some point by CMake upstream (https://gitlab.kitware.com/cmake/cmake/-/issues/20479).

JensWehner:
can he not simply specify the paths to the Eigen installation?

junghans:
> can he not simply specify the paths to the Eigen installation?

I guess, the admin just copied the files over and didn't actually build Eigen3 with CMake hence `Eigen3Config.cmake` is missing in the install.

JensWehner:
yes, seems like it.

jkrajniak:
It seems like Eigen on the cluster was installed via easybuild (I see only easybuild configs installed). From I see here, this solution could be difficult to maintain, so I will close PR.

PR #653, opened 2021-02-11
Maps Switch to unique_ptrs

JoshuaSBrown:
Part 3


include/votca/csg/topologymap.h
```diff
@@ -30,26 +31,26 @@ namespace csg {
...
+  void AddMoleculeMap(std::unique_ptr<Map> map);
 
   void Apply();
 
  private:
   Topology *_in;
   Topology *_out;
 
-  using MapContainer = std::vector<Map *>;
+  using MapContainer = std::vector<std::unique_ptr<Map>>;
```
JensWehner: why not just make it

```suggestion
  using MapContainer = std::vector<Map>;
```

JensWehner: and make it value based.

JensWehner (changes requested):
either use `make_unique` or preferably make it value based and not pointer based

junghans:
@JoshuaSBrown 
```
/__w/csg/csg/votca/csg/include/votca/csg/topologymap.h:52:33: error: no matching function for call to 'std::vector<votca::csg::Map>::push_back(std::remove_reference<std::unique_ptr<votca::csg::Map>&>::type)'
   52 |   _maps.push_back(std::move(map));
      |                                 ^
```

junghans:
No app using the mapping fail:
```
	 92 - integration_Run_csg_fmatch (SEGFAULT)
	 94 - integration_Run_csg_fmatch_3body (SEGFAULT)
	100 - integration_Run_csg_map_gro (SEGFAULT)
```

JoshuaSBrown:
Might need to wait until the topology map pr gets merged before this will work.

include/votca/csg/map.h
```diff
@@ -97,6 +86,8 @@ class Map_Sphere : public BeadMap {
     double _force_weight;
   };
   std::vector<element_t> _matrix;
+
+  friend class Map;
```
JensWehner: I do not understand.

JensWehner (changes requested):
I do not understand the hiding of all the constructors. I understand that you want to protect certain conditions, but in my mind that is what assert statements are for. I do not want to couple constructors to other classes, even less friend them, because then encapsulation goes out of the window. 

If you want to create a common interface for object creation, I would make a factory.

JoshuaSBrown:
> I do not understand the hiding of all the constructors. I understand that you want to protect certain conditions, but in my mind that is what assert statements are for. I do not want to couple constructors to other classes, even less friend them, because then encapsulation goes out of the window.
> 
> If you want to create a common interface for object creation, I would make a factory.

Ok, I think I can make that work, I'll just move all the child classes into the source files, that should hide them from being used where they are not supposed to.

junghans:
@votca-bot format

JensWehner (approved):
thanks

PR #655, opened 2021-02-11
Change residues to unique_ptr Fix memory management

JoshuaSBrown:
Part 5


include/votca/csg/topology.h
```diff
@@ -46,7 +47,7 @@ class Interaction;
 
 using MoleculeContainer = std::vector<Molecule *>;
 using BeadContainer = std::vector<Bead *>;
-using ResidueContainer = std::vector<Residue *>;
+using ResidueContainer = std::vector<std::unique_ptr<Residue>>;
```
JensWehner: can we not make them value based, i.e. std::vector<Residue>

include/votca/csg/topology.h
```diff
@@ -453,15 +455,13 @@ inline Molecule *Topology::CreateMolecule(std::string name) {
...
-  return res;
+  _residues.emplace_back(new Residue(id, name));
+  return _residues.back().get();
 }
 
 inline Residue *Topology::CreateResidue(std::string name) {
-  Residue *res = new Residue(_residues.size(), name);
-  _residues.push_back(res);
-  return res;
+  _residues.emplace_back(new Residue(_residues.size(), name));
```
JensWehner: we have the ownership problem here again.

JensWehner (commented):
Good changes, please use more `make_unique` otherwise 

I would prefer CSG to be more value based than pointer based, but I imagine is difficult to entangle all this, so please ignore my comments if not helpful.

include/votca/csg/topology.h
JoshuaSBrown: What do you mean? _residues is a vector of unique_ptrs it owns the residue?

JensWehner: okay can you use `make_unqiue<Residue>(size,name)` then, there should not be any `new`in the code

JoshuaSBrown:
Do not merge yet, the vector needs to be changed. Because pointers to the residues will be invalidated if the size of the vector changes.

JoshuaSBrown:
> what is the reason behind this change?

What do you mean, which changes are you referring to? I think I made several since you last looked. I essentially got rid of indirection of pointers and placed Residues directly in a vector. This has the advantage that it is more performant because you do not have the indirection associated with objects created on the heap. However, the problem with the std vector is that it does not guarantee your objects will remain at the same address if the vector changes size. This means that if you are using pointers that point to objects stored in your object you may very well invalidate them. The boost stable_vector prevents this, it gaurantees that your address remains sane no matter how much you grow. It does this from what I remember by not enforcing a continuity between every element in the vector.

JensWehner:
Hi, 

yes I meant the stable::vector change. The problem with stable::vector is that it is basically an indexed-linked list under the hood, so there is no performance benefit from it.

JensWehner (approved):

JoshuaSBrown:
> Hi,
> 
> yes I meant the stable::vector change. The problem with stable::vector is that it is basically an indexed-linked list under the hood, so there is no performance benefit from it.

Oh, I don't think I realized this, I thought it was a linked list of vectors, where you have groups of continuous memory. Do you know if such a structure exists?

JensWehner:
I think std::deque has pointer stability and sth. along those lines. 

https://en.cppreference.com/w/cpp/container/deque

JoshuaSBrown:
> I think std::deque has pointer stability and sth. along those lines.
> 
> https://en.cppreference.com/w/cpp/container/deque

Well do you have an opinion on what would be better to use?

JensWehner:
I would probably go for std::deque then. Maybe performance does not matter at all. I do not know.

JoshuaSBrown:
95: Test command: /home/jbrown/Software/csg/build/src/tests/unit_test_performance
95: Test timeout computed to be: 10000000
95: Running 1 test case...
95: Testing push_back stable vector with int
95:  0.005968s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing push_back stable vector with large structure
95:  0.013078s wall, 0.020000s user + 0.000000s system = 0.020000s CPU (152.9%)
95: Testing access stable vector with int
95:  0.000486s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing access stable vector with large structure
95:  0.001390s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing push_back deque with int
95:  0.000184s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing push_back deque with large structure
95:  0.005663s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing access deque with int
95:  0.000156s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing access deque with large structure
95:  0.001332s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)

JoshuaSBrown:
Testing range based for with large structure for stable_vector
95:  0.000723s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
Testing range based for with large structure for stable_vector
95:  0.000763s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)

JoshuaSBrown:
95: Testing range based for with large structure for stable_vector
95:  0.000452s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing range based for with large structure for deque
95:  0.000451s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: Testing range based for with large structure for boost deque with initially allocated storage
95: Size of large object is 104
95: block size 104  0.000358s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: block size 208  0.000309s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: block size 416  0.000273s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)
95: block size 832  0.000264s wall, 0.000000s user + 0.000000s system = 0.000000s CPU (n/a%)

So it looks like the boost deque is actually probably the best way to go. Unfortunately, the std::deque does not allow you to customize the block size it seems to be compiler and system dependent how that happens. For the large structures that are stored in the topology object it would appear that the performance benefits of a standard deque are lost because the default block size is too small. Of course if the block size is too large you end up wasting memory in the allocation, but you can already see for fitting 4 objects you get about double the speedup of the std deque and stable vectors at least for range based for loop iteration. 

Now knowing what I know now, I will probably always avoid using std::deque moving forward, they should have let you specify the block size on construction.

JoshuaSBrown:
So moving forward, I will change the container to a boost deque with block size equal to 4* the class size.

junghans:

JoshuaSBrown:
So what will you do to if I say no. 😜

JensWehner:
@votca-bot format

JoshuaSBrown:
@junghans this is blocking for no reason.

junghans:
@JoshuaSBrown huh?

PR #656, opened 2021-02-11
Joshua s brown/bead to unique ptr

JoshuaSBrown:
Memory issues part 6.


include/votca/csg/topology.h
```diff
@@ -441,9 +441,9 @@ inline Bead *Topology::CreateBead(Bead::Symmetry symmetry, std::string name,
                                   std::string type, Index resnr, double m,
                                   double q) {
 
-  Bead *b = new Bead(_beads.size(), type, symmetry, name, resnr, m, q);
-  _beads.push_back(b);
-  return b;
+  _beads.emplace_back(
+      new Bead(_beads.size(), type, symmetry, name, resnr, m, q));
```
JensWehner: You make a copy here now. Before they just copied the pointer, I think?

src/libcsg/modules/io/pdbwriter.cc
```diff
@@ -73,7 +73,7 @@ void PDBWriter::WriteBox(const Eigen::Matrix3d &box) {
   _out << boxfrmt % a % b % c % alpha % beta % gamma;
 }
 
-void PDBWriter::writeSymmetry(Bead *bead) {
+void PDBWriter::writeSymmetry(const std::unique_ptr<Bead> &bead) {
```
JensWehner: I would prefer just a `const Bead& bead`

JensWehner (commented):
good changes, ignore the value stuff if it complicates things.

include/votca/csg/topology.h
JoshuaSBrown: I'm confused are we looking at the same lines of code :). They were calling `new Bead` so ya they were storing the pointer in the vector but the memory was on the heap and that vector is responsible for the memory management of the bead.

src/libcsg/modules/io/pdbwriter.cc
JoshuaSBrown: Ya but that does not work well with the vector of unique pointers stored in the topology class.

JoshuaSBrown: You would have to do something special to get the begin and end iterator methods to work like that. If you can figure out how I will implement it though.

include/votca/csg/topology.h
JensWehner: I did not look at the line above. YOu are right.

src/libcsg/modules/io/pdbwriter.cc
JensWehner: you can just call `writeSymmetry(*bead)` then and dereference the pointer.

JensWehner: which call of this function do you think is problematic?

include/votca/csg/pdbwriter.h
```diff
@@ -104,7 +108,7 @@ inline void PDBWriter::WriteContainer(T &container) {
   boost::format atomfrmt(
       "ATOM  %1$5d %2$-4s %3$-3s %4$1s%5$4d    %6$8.3f%7$8.3f%8$8.3f\n");
 
-  for (auto &atom : getIterable(container)) {
+  for (const auto &atom : getIterable(container)) {
```
JoshuaSBrown, line 106: Jens this is the reason I pass in a unique_ptr to writeSymmetry() the beads are now stored as unique pointers in topology. When getIterable is called it is by default getting the begin and end bethods of the _beads container in topology. This means atom will be of type unique_ptr. I would have to store another container just with Bead * to do what you are asking unless I'm missing something?

JensWehner: okay

junghans:
ping @JensWehner

include/votca/csg/topology.h
```diff
@@ -46,7 +49,9 @@ namespace csg {
 class Interaction;
 
 using MoleculeContainer = std::vector<Molecule *>;
-using BeadContainer = std::vector<Bead *>;
+// Stable vectors are necessary to be sure that the Bead pointers that point
+// to addresses in the vector are not invalidated as the vector grows.
```
JensWehner: I mean the other way to do it is just build the vector of beads first afterwards it is stable.

include/votca/csg/topology.h
```diff
@@ -423,7 +429,7 @@ class Topology {
 
   std::map<std::string, Index> _interaction_groups;
 
-  std::map<std::string, std::list<Interaction *> > _interactions_by_group;
+  std::map<std::string, std::list<Interaction *>> _interactions_by_group;
```
JensWehner: can we make this 
```suggestion
  std::map<std::string, std::vector<Interaction *>> _interactions_by_group;
```

JensWehner (approved):
Looking good, we should make sure, that stable::vector does not kill performance, otherwise nice changes

junghans:
@JensWehner @JoshuaSBrown:
```
/__w/csg/csg/votca/csg/src/libcsg/topology.cc:265:46: error: no viable conversion from '_Rb_tree_iterator<pair<[...], std::vector<votca::csg::Interaction *, std::allocator<votca::csg::Interaction *>>>>' to '_Rb_tree_iterator<pair<[...], std::__cxx11::list<votca::csg::Interaction *, std::allocator<votca::csg::Interaction *>>>>'
  map<string, list<Interaction *>>::iterator iter =
                                             ^
```

junghans:

junghans:
@votca-bot format

junghans:
@JoshuaSBrown more conflicts.

junghans:
@votca-bot format

PR #658, opened 2021-02-12
Joshua s brown/interactions to unique ptr

JoshuaSBrown:
Improved encapsulation of interaction class added const correctness and improved memory management.


include/votca/csg/interaction.h
```diff
@@ -100,224 +113,62 @@ class Interaction {
...
-  double EvaluateVar(const Topology &top) override;
-  Eigen::Vector3d Grad(const Topology &top, Index bead) override;
+  virtual ~IDihedral() = default;
+  double EvaluateVar(const Topology &top) const override;
+  Eigen::Vector3d Grad(const Topology &top, Index bead) const override;
 
  private:
-};
-
-inline double IBond::EvaluateVar(const Topology &top) {
```
JensWehner, line 198: ohh thanks for taking away all the inline stuff.

include/votca/csg/molecule.h
```diff
@@ -97,7 +99,9 @@ class Molecule {
 inline Index Molecule::getBeadIdByName(const std::string &name) {
   Index i = getBeadByName(name);
   if (i < 0) {
-    { return i; }
+    {
```
JensWehner: I do not understand the second layer of brackets.

include/votca/csg/topology.h
```diff
@@ -105,6 +108,36 @@ class Topology {
   Residue *CreateResidue(std::string name);
   Residue *CreateResidue(std::string name, Index id);
 
+  // Should not be able to alter interactions from outside of topology
+  template <class T>
+  const Interaction *CreateInteraction(T bead_ptrs, const std::string &grp,
+                                       const Index ind, const Index mol_ind) {
+    if (bead_ptrs.size() == 2) {
+      _interactions.push_back(std::unique_ptr<IBond>(new IBond(bead_ptrs)));
```
JensWehner:
```suggestion
      _interactions.push_back(std::make_unique<IBond>(bead_ptrs));
```

include/votca/csg/topology.h
```diff
@@ -184,8 +217,8 @@ class Topology {
     return _interactions;
   }
 
-  void AddBondedInteraction(Interaction *ic);
-  std::list<Interaction *> InteractionsInGroup(const std::string &group);
+  std::list<const Interaction *> InteractionsInGroup(
```
JensWehner: Would a std::vector not be better?

include/votca/csg/topology.h
```diff
@@ -426,7 +459,7 @@ class Topology {
 
   std::map<std::string, Index> _interaction_groups;
 
-  std::map<std::string, std::list<Interaction *> > _interactions_by_group;
+  std::map<std::string, std::list<const Interaction *>> _interactions_by_group;
```
JensWehner: same here. I think up to 50000 `int` or pointers std::vector is much faster than list

src/csgapps/partial_rdf/rdf_calculator.cc
```diff
@@ -298,9 +298,9 @@ void RDFCalculator::Worker::DoBonded(Topology *top) {
     _current_hists[i._index].Clear();
 
     // now fill with new data
-    std::list<Interaction *> list = top->InteractionsInGroup(name);
+    std::list<const Interaction *> list = top->InteractionsInGroup(name);
 
-    for (auto ic : list) {
+    for (const auto &ic : list) {
```
JensWehner:
```suggestion
    for (const auto &ic : top->InteractionsInGroup(name)) {
```

src/libcsg/cgmoleculedef.cc
```diff
@@ -175,22 +175,8 @@ Molecule *CGMoleculeDef::CreateMolecule(Topology &top) {
...
-      } else {
-        throw runtime_error("unknown bonded type in map: " + prop->name());
-      }
-
-      ic->setGroup(iagroup);
-      ic->setIndex(index);
-      ic->setMolecule(minfo->getId());
-      top.AddBondedInteraction(ic);
+      const Interaction *ic =
+          top.CreateInteraction(atoms, iagroup, index, minfo->getId());
```
JensWehner, line 180: nice

src/libcsg/interaction.cc
```diff
@@ -0,0 +1,171 @@
...
+
+namespace votca {
+namespace csg {
+
+using namespace std;
+
+void Interaction::RebuildName() {
+  std::stringstream s;
+  if (_mol != -1) {
+    {
```
JensWehner: I think something in the formatter is wrong or so

src/libcsg/interaction.cc
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+  if (!_group.empty()) {
+    s << ":" << _group;
+    if (_group_id != -1) {
+      s << " " << _group_id;
+    }
+  }
+  if (_index != -1) {
+    {
```
JensWehner: here again

src/libcsg/modules/io/dlpolytopologyreader.cc
```diff
@@ -302,34 +302,33 @@ bool DLPOLYTopologyReader::ReadTopology(string file, Topology &top) {
...
                          // are merely skipped (ignored)
             Index ids[4];
-            Interaction *ic = nullptr;
             sl >> ids[0];
             sl >> ids[1];
+            std::list<Index> bead_ids;
             if (type == "BONDS") {
-              ic = new IBond(id_map[ids[0] - 1],
-                             id_map[ids[1] - 1]);  // -1 due to fortran vs c
+              bead_ids.push_back(id_map[ids[0] - 1]);
```
JensWehner, line 309: nice

src/libcsg/topology.cc
```diff
@@ -261,25 +256,12 @@ void Topology::CheckMoleculeNaming(void) {
...
-
-std::list<Interaction *> Topology::InteractionsInGroup(const string &group) {
-  map<string, list<Interaction *>>::iterator iter =
+std::list<const Interaction *> Topology::InteractionsInGroup(
+    const string &group) const {
+  map<string, list<const Interaction *>>::const_iterator iter =
       _interactions_by_group.find(group);
   if (iter == _interactions_by_group.end()) {
-    return list<Interaction *>();
+    return list<const Interaction *>();
```
JensWehner: I would prefer std::vector simply for performance reasons but this PR is valuable in its current form

src/tools/csg_dlptopol.cc
```diff
@@ -245,7 +245,7 @@ void DLPTopolApp::WriteMoleculeInteractions(ostream &out, Molecule &cg) {
   votca::Index n_entries = 0;
   votca::Index nb = -1;
 
-  for (Interaction *ic : cg.Interactions()) {
+  for (const auto &ic : cg.Interactions()) {
```
JensWehner: ic is not a good name, especially with the `auto`

JensWehner: but it is common in csg. So this is a nitpick

JensWehner (approved):
There are a couple of more make_unique which I consider good practice, otherwise 
there is std::list vs std::vector which is a nitpick,
https://baptiste-wicht.com/posts/2012/11/cpp-benchmark-vector-vs-list.html

include/votca/csg/topology.h
JoshuaSBrown: Ok I'll try to make the switch.

include/votca/csg/molecule.h
JoshuaSBrown: Ya, good catch, don't know how that happened.

JoshuaSBrown: Might have been clang formatting...

src/tools/csg_dlptopol.cc
JoshuaSBrown: Actually I think ic is short for interaction container so I should really change it.

include/votca/csg/interaction.h
```diff
@@ -25,71 +25,89 @@
...
-  virtual Eigen::Vector3d Grad(const Topology &top, Index bead) = 0;
+  virtual Eigen::Vector3d Grad(const Topology &top, Index bead) const = 0;
   Index BeadCount() const { return _beads.size(); }
   Index getBeadId(Index bead) const {
     assert(bead > -1 && boost::lexical_cast<size_t>(bead) < _beads.size());
     return _beads[bead];
   }
 
  protected:
+  void setIndex(const Index &index) {
```
JensWehner, line 82: What is the intention of hiding these methods?

include/votca/csg/interaction.h
```diff
@@ -100,224 +118,62 @@ class Interaction {
...
-    }
-  }
-  double EvaluateVar(const Topology &top) override;
-  Eigen::Vector3d Grad(const Topology &top, Index bead) override;
+  virtual ~IBond() = default;
+  double EvaluateVar(const Topology &top) const override;
+  Eigen::Vector3d Grad(const Topology &top, Index bead) const override;
 
  private:
+  template <class T>
```
JensWehner, line 131: Why hiding the contructor?

JensWehner: This couples the interaction to topology.

include/votca/csg/interaction.h
```diff
@@ -100,224 +118,62 @@ class Interaction {
...
-             ((v1 + v2) / (v1.norm() * v2.norm()) -
-              (v1.dot(v2)) * (v2.squaredNorm() * v1 + v1.squaredNorm() * v2) /
-                  (std::pow(v1.norm(), 3) * std::pow(v2.norm(), 3)));
-      break;
-    case (2):
-      return acos_prime * (-v1 / (v1.norm() * v2.norm())) +
-             (v1.dot(v2) * v2 / (v1.norm() * std::pow(v2.norm(), 3)));
-      break;
+  template <class T>
+  IDihedral(T &beads) {
```
JensWehner, line 169: I do not understand.

include/votca/csg/molecule.h
```diff
@@ -70,14 +70,16 @@ class Molecule {
   std::string getBeadName(Index bead) { return _bead_names[bead]; }
 
   /// Add an interaction to the molecule
-  void AddInteraction(Interaction *ic) { _interactions.push_back(ic); }
+  void AddInteraction(const Interaction *ic) { _interactions.push_back(ic); }
```
JensWehner, line 77: Who in general has the ownership of interaction topology/residues or what?

src/tests/test_interaction.cc
```diff
@@ -24,82 +24,84 @@
...
 BOOST_AUTO_TEST_CASE(test_interaction_setters_getters) {
 
-  IBond bond1(1, 2);
-  bond1.setGroup("large");
-  bond1.setGroupId(1);
-  bond1.setIndex(1);
-  bond1.setMolecule(1);
+  Topology top;
+  const Interaction* bond1 =
+      top.CreateInteraction(std::list<votca::Index>{1, 2}, "large", 1, 1);
```
JensWehner, line 55: Why can I not create Interactions without a topology? I do not think it makes the code more  flexible.

JoshuaSBrown:
1. Encapsulation normally comes at the cost of flexibility.
2. In csg, the topology class manages ownership of all of the topology objects (bead, molecule, residues, interactions).
3. The Interaction class was the only topology type object not owned by the topology class this pr makes control all of the topology objects (bead, residue, molecule, interaction) consistent. 
4. The interaction methods that have been made protected with the exception of the friend class are not used outside of the topology object anyway. The flexibility of having public setters is not needed. Furthermore, if the previously public setters were used outside of the topology class then it could cause problems because the "interaction groups" members in the topology class depend on how the values of each interaction are defined. Making the topology class a friend class makes explicit this relationship.

JensWehner:
I agree that managing the ownership in topology is a good idea, but I do not think only allowing topology to create these objects is a good idea. What are the benefits of that?

In my mind it makes writing unit tests more complicated
It gives the wrong message because an interaction can exist without  topology. 

If you do not need certain functions do not hide them, just delete them. 

I do not like friend classes at all, I think there should be none. It just allows you to add more connections which should not exist.

include/votca/csg/interaction.h
JoshuaSBrown: It reduces scope, better encapsulation.

include/votca/csg/interaction.h
JoshuaSBrown: Yes, and makes it more consistent with the design. I already responded in slack, so we can go from there.

include/votca/csg/molecule.h
JoshuaSBrown: Topology has ownership of all the topology objects (beads, residues, interactions, molecules).

src/tests/test_interaction.cc
JoshuaSBrown: No, it does not make it more flexible, it reduces scope. In this case flexibility is not needed. If you look into the topology object CreateInteraction method you can get some insight as to why you would not want to change the interaction outside of the topology.

JoshuaSBrown:
> I agree that managing the ownership in topology is a good idea, but I do not think only allowing topology to create these objects is a good idea. What are the benefits of that?

I think I have already answered that. 
> 
> In my mind it makes writing unit tests more complicated

Perhaps
> It gives the wrong message because an interaction can exist without topology.

What wrong impression is that? Does it make sense to have interactions when there are no beads or molecules all of which are topology objects? 

> If you do not need certain functions do not hide them, just delete them.

This is a valid point the set methods are not really needed sense the topology object has been made a friend. 

> I do not like friend classes at all, I think there should be none. It just allows you to add more connections which should not exist.

This is a preference, I think, which hold to the purity of object oriented design principles. I think for the most part it is good to hold object oriented principles, however I will point out I did not create this design I have just made it consistent, secondly I do not think the design is damaging or bad. In this case you sacrifice the flexibility, but this does not really matter in this case because you don't need it. You lose encapsulation of the interaction class in terms of how the topology class interacts withit, however you consider the interaction and topology classes as a unit you actually have better encapsulation because you can minimize the scope of how the interaction classes can be changed outside of the topology class.

JoshuaSBrown:
I saw this in stack overflow and I think it falls along the design logic used here:

```
C++ friends exist just for that. Every other surrogate will simply mimic the same concept.

* There are thing that are for everyone,
* Things for you only, (eventually inherited from your children)
* Things shared by you with a well define and limited set of people.

Those "people" are called "friend". Eliminating that concept will require those things to become public, or to become private, but with public accessors to be managed. You must also expect anyone to be capable of doing things with them out of your control.
```
Emilio Garavaglia

JoshuaSBrown:
@JensWehner There is an improvement to this approach so that the topology object does not have access to all the internals of the interaction class. Look here, I would be in favor of doing something
 similar. https://stackoverflow.com/questions/2902717/c-is-there-a-way-to-limit-access-to-certain-methods-to-certain-classes-withou. Though I'm not sure this would work with a constructor.

include/votca/csg/interaction.h
JoshuaSBrown: I wish there was a way to only expose some methods to certain classes.

include/votca/csg/interaction.h
JoshuaSBrown: This is the same argument.

JensWehner:
Okay I understand your point, my attitude is that I am happily accept dangling pointers if someones uses a class in a way which it is not supposed to be, whereas you want to make it safe and reduce exactly these problems for the price of loosing some OO. You have good arguments, I do not agree with. 

What I certainly agree with is that your changes improve the code to before. So thanks for the good work and I approve this.

JoshuaSBrown:
@junghans what do I need to do to build integration_Run_csg_boltzmann_ppy? When I build locally the test is not found.

junghans:
@JoshuaSBrown You need to have gromacs (library devel package and `gmx` binary`) installed!

junghans:

junghans:
@JoshuaSBrown this one can go next.

junghans:
ping @JoshuaSBrown

junghans:
@JoshuaSBrown Ok it seems the ppy test is still broken
``
2021-02-28T12:53:28.9752030Z         Start 157: integration_Run_csg_boltzmann_ppy
2021-02-28T13:14:06.6562387Z  86/157 Test #157: integration_Run_csg_boltzmann_ppy ..................Child killed***Exception: 1237.22 sec
2021-02-28T13:14:06.6563617Z Reading file topol.tpr, VERSION 2019.6 (double precision)
2021-02-28T13:14:06.6564388Z I have 82 beads in 1 molecules
2021-02-28T13:14:06.6565089Z 
```

junghans:
ping @JoshuaSBrown

JoshuaSBrown:
ping @junghans 

Hey you forgot.

junghans:
@JoshuaSBrown How can I help?

JoshuaSBrown:
> @JoshuaSBrown How can I help?

Create some time. :)

JoshuaSBrown:
So I found one of the problems. However, I could use some help understanding what csg_boltzmann_ppy is doing?

JoshuaSBrown:
What exactly is supposed to be in excl.txt?

junghans:
> What exactly is supposed to be in excl.txt?

The exclusions between the atoms are in `excl.txt`, see https://www.votca.org/csg/methods.html#generating-exclusion-lists.

JoshuaSBrown:
2 problems,

there is a while loop that was removing beads as it was iterating through them
secondly, in my refactor I failed to account for non-bonded interactions.

junghans:
Replace by https://github.com/votca/votca/pull/815

PR #662, opened 2021-02-16
Fix memory leak in splines

JoshuaSBrown:
Removes the use of new with Splines, this was causing a memory leak.


JensWehner (approved):
Nice! Just really nice

PR #691, opened 2021-04-30
Drop gmx2020

junghans:


junghans:
@votca-bot format

junghans:
@votca-bot changelog: format code with clang-12 and drop gmx2020 builds

PR #693, opened 2021-05-10
changed tokenizer interface

JensWehner:


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: adapted tokenizer api

src/libcsg/modules/io/lammpsdumpreader.cc
```diff
@@ -161,7 +158,7 @@ void LAMMPSDumpReader::ReadAtoms(Topology &top, string itemline) {
 
   {
     tools::Tokenizer tok(itemline.substr(12), " ");
-    tok.ToVector(fields);
+    fields = tok.ToVector();
```
felipeZ, line 161: should this `field` assignment be written in a single line together with the `fields` declaration?

felipeZ (approved):

src/libcsg/modules/io/lammpsdumpreader.cc
JensWehner: yeah fields is declared earlier. So I only wanted to change the tokenizer api and not everything else. I agree in principle, but I did not want to do a larger scale refactoring.

PR #697, opened 2021-07-08
fix pos scaling (ang2nm) in lammps data reader

junghans:
Report on mailing list: https://groups.google.com/g/votca/c/C3L0xhdUs6s/m/4qg33QS8AgAJ

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ Version 2021.1 (released XX.03.21)
 ==================================
 
 -  format code with clang-12 and drop gmx2020 builds (#691)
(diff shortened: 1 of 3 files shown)


JensWehner (approved):
I am okay with that, 
the problem to some extend still is that LAMMPS is unit agnostic to some extend, so it is really hard to tell which units it is actually using in the computation.

JensWehner:
@votca-bot format

junghans:
> I am okay with that,
> the problem to some extend still is that LAMMPS is unit agnostic to some extend, so it is really hard to tell which units it is actually using in the computation.

True, but be scale the box size in the same reader!

PR #703, opened 2021-07-30
Lammps reader

JensWehner:
This PR does two things:

a) throw an error when votca tries to construct a bond between atoms from different molecules in the lammps reader.

b) at the moment unnamed molecules are named by their chemical composition, this can be problematic if we have isomers in the topology in that case an index is added at the end of the molecule.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,7 @@ Version 2022-dev
 -  adapted tokenizer api (#693)
 -  made membervariable format consistent (#694)
 -  removed unused functions (#702)
+-  reworked lammps molecule naming (#703)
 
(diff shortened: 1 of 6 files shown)


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: reworked lammps molecule naming

junghans:
This is more @JoshuaSBrown's cup of tea.

JensWehner:
It is fully backwards compatible it just detects more errors than before.

junghans (approved):
Can you add a test?

JensWehner:
Yes I will next week.

JensWehner:
@votca-bot format

